PR #2, opened 2022-01-08
Audio param setup with generics and Any-trait downcasting - do not merge this!

orottier:
Hey I tried to follow along with the generics path for the audio params, it was a disaster.
I need to erase the generic params by Boxing them and wrapping them in a special AnyTrait. Otherwise it would not play nicely with napi.
In any case, this is for your entertainment only. Do not merge it, the enum-way is better!


b-ma:
Hey, thanks for having a look and doing the exercise, I'm not sure I understand everything going on there (I'm quite sure I don't actually :)... but ok... let's go for the brute force solution then!

Issue #4, opened 2022-10-24
Binding OfflineAudioContext


bthj:
Hi - super interesting project you have here!

It seems that `OfflineAudioContext` hasn't yet been mapped ([web-audio-api-rs](https://github.com/orottier/web-audio-api-rs) seems to implement it).  If I'd try to add that binding, should it be something similar to what has already been done for `AudioContext`?
https://github.com/ircam-ismm/node-web-audio-api/blob/main/index.js#L91

b-ma:
Hi, thanks!

No indeed, still have some work to reach the `web-audio-api-rs` current state. Actually, implementing the bindings for the offline context should not be much complicated I think. Starting from https://github.com/ircam-ismm/node-web-audio-api/blob/main/generator/templates/audio_context.tmpl.rs it should be mostly duplicating the file, modifying the constructor to handle the constructor signature and removing the extra online context methods (resume, suspend, etc.)

The most complicated thing to grasp I guess is that all the rust source files for the bindings located in `src` are dynamically generated from the files and scripts located in the `generator` directory, so the general workflow is:
- create / modify the files in `generator`
- generate the rust source files from the templates with `npm run generate`
- build the native node module from the generated rust sources with`npm run build`

Let me know if you want to have a try and/or need some help, I can also have a shot on this if you don't feel confortable with the source code.

bthj:
Thanks for the insights.  Maybe I should have a try and see how far I get.  Will post back :)

b-ma:
ok cool, let me know!

bthj:
I've taken a stab at adding this but it's not quite building yet.  Here's what I've done so far:

https://github.com/ircam-ismm/node-web-audio-api/compare/main...bthj:node-web-audio-api:feature/offline_audio_context

The constructor signature is different (offline with 3 arguments while online with one options arg) and how that's handled doesn't seem to produce errors.  But an attempt at implementing a binding for the `start_rendering` results in errors.  There I started by copying the implementation for `decode_audio_data` as it's also handling an audio buffer, but the signature is a bit different, where `decode_audio_data` returns a `Result<Audiobuffer ... >` but `start_rendering` returns `AudioBuffer` - and that's causing a build error I currently don't know how to handle due to unfamiliarity with napi-rs.  It says 
`cannot use the '?' operator in a function that returns 'AudioBuffer'`
at 
`let napi_obj = ctx.env.unwrap::<NapiOfflineAudioContext>(&js_this)?;`

[...]

PR #5, opened 2022-10-31
Steps towards implementing bindings for the offline audio context 

bthj:
- does not build.

See:  https://github.com/ircam-ismm/node-web-audio-api/issues/4

### DIFF
--- a/generator/templates/offline_audio_context.tmpl.rs
+++ b/generator/templates/offline_audio_context.tmpl.rs
@@ -1,0 +1,296 @@
+use std::fs::File;
+use napi::*;
+use napi_derive::js_function;
+use web_audio_api::context::*;
+use crate::*;
+use web_audio_api::AudioBuffer;
+
+
+pub(crate) struct NapiOfflineAudioContext(OfflineAudioContext);
+
+impl NapiOfflineAudioContext {
+    pub fn create_js_class(env: &Env) -> Result<JsFunction> {
+        env.define_class(
+            "OfflineAudioContext",
+            constructor,
+            &[
+                Property::new("Symbol.toStringTag")?
+                    .with_value(&env.create_string("OfflineAudioContext")?)
+                    .with_property_attributes(PropertyAttributes::Static),
+
+                Property::new("currentTime")?.with_getter(get_current_time),
+                Property::new("sampleRate")?.with_getter(get_sample_rate),
+                Property::new("state")?.with_getter(get_state),
+                // Property::new("baseLatency")?.with_getter(get_base_latency),
+                // Property::new("outputLatency")?.with_getter(get_output_latency),
+
+                // for now async methods are sync, from a JS perpspective the
+                // API will nonetheless be the same... (see monkey-patch.js)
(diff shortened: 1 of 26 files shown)


b-ma:
There is still a weird SegFault when the script exits (or maybe when the node is garbage collected which would be more annoying...), but I don't really how to fix that right now

Added a simple example in `examples/offline.mjs`

generator/templates/offline_audio_context.tmpl.rs
```diff
@@ -0,0 +1,301 @@
+
+use std::fs::File;
+use napi::*;
+use napi_derive::js_function;
+use web_audio_api::context::*;
+use crate::*;
+// use web_audio_api::AudioBuffer;
+
+
+pub(crate) struct NapiOfflineAudioContext(OfflineAudioContext);
```
orottier, line 10: I realize that `start_rendering` consuming `self` may be fundamentally incompatible with JS way of working.
You could try to redefine the struct as
`struct NapiOfflineAudioContext(RefCell<Option<OfflineAudioContext>>)`
and when `start_rendering` is called, take the context out and leave `None` in place.
In this case I do not think you need to resort to `unsafe` code, which I presume causes a double drop of the context (hence segfault)

orottier: Alternatively, we could expose a non-consuming version of the `startRendering` method and hide it behind an undocumented feature flag in the parent library

b-ma: Thank for the ideas! 
I think we should really try the first solution before considering adding special things to the `rs` lib.

I only wonder what should vs what actually happen if we do something like:
```rs
let context = OfflineAudioContext::new();
// later
let buffer = context.start_rendering();
// later
context.decode_audio_data()
```
(agreed that it would be quite weird code)

b-ma: Seems to work with a simple `struct NapiOfflineAudioContext(Option<OfflineAudioContext>)`, thanks!

orottier: Great!

> I only wonder what should vs what actually happen if we do something like:

I think this is our punishment for deviating from the spec.
We could still roll back the change and make startRendering take `&self`.
In that case, it should return an `Result<AudioBuffer>` because we should throw an `InvalidStateError` when rendering has already started

b-ma: > In that case, it should return an Result<AudioBuffer> because we should throw an InvalidStateError when rendering has already started

Yup, no strong opinion on my side, in one way that would make the API aligned with `decode_audio_audio` which is not really chocking, on the other way it would complicate a bit the user code. In any case, this would be very easy to change here if needed

b-ma:
@bthj do you want to add/review something? or should I merge?

bthj:
@b-ma it would be great if you'd merge.  Then I'll test this soon.  Thanks.

b-ma:
Ok, thanks for pushing this forward!

Issue #9, opened 2022-11-07
OfflineAudioContext constructor with options object


bthj:
To have the [OfflineAudioContext constructor](https://developer.mozilla.org/en-US/docs/Web/API/OfflineAudioContext/OfflineAudioContext) accept either an `options` object or a set of parameters, I made a small modification here:

https://github.com/ircam-ismm/node-web-audio-api/compare/main...bthj:node-web-audio-api:feature/offline_audio_context_options_constructor

Should I open a pull request?

orottier:
> maybe we could handle it at the rust level

Not in any elegant way. We cannot overload the `new` method so we would need to introduce `new_from_options`. I don't think it will give any benefit

bthj:
Here's a PR for one way of handling this in JS: https://github.com/ircam-ismm/node-web-audio-api/pull/11

b-ma:
fixed in https://github.com/ircam-ismm/node-web-audio-api/pull/11

PR #11, opened 2022-11-08
Feature/offline audio context options constructor

bthj:
See:  https://github.com/ircam-ismm/node-web-audio-api/issues/9


monkey-patch.js
```diff
@@ -88,8 +88,16 @@ function patchAudioContext(NativeAudioContext) {
 function patchOfflineAudioContext(NativeOfflineAudioContext) {
   class OfflineAudioContext extends NativeOfflineAudioContext {
     constructor(...args) {
-      super(...args);
-      console.log(args);
+      // handle initialisation with either an options object or a sequence of parameters
+      // https://developer.mozilla.org/en-US/docs/Web/API/OfflineAudioContext/OfflineAudioContext#parameters
```
b-ma, line 92: could you replace the link with this one https://webaudio.github.io/web-audio-api/#dom-offlineaudiocontext-constructor-contextoptions-contextoptions? mdn is very nice, but the spec in more precise

monkey-patch.js
```diff
@@ -88,8 +88,16 @@ function patchAudioContext(NativeAudioContext) {
...
+      if( typeof args[0] === 'object'
+          && 'numberOfChannels' in args[0] && 'length' in args[0] && 'sampleRate' in args[0]
```
b-ma, line 94: Would be nice to check each of the parameter and throw a NotSupportedError if something is not right
> A [NotSupportedError](https://webidl.spec.whatwg.org/#notsupportederror) exception MUST be thrown if any of the arguments is negative, zero, or outside its nominal range.

Error handling is still in the todo list on the rust side, but we will have in any case something crashing with `NumberExpected` which is not very informative, if we do something like:
```
const offline = new OfflineAudioContext({
  sampleRate: 0,
});
```

bthj: Now throwing that error if the constructor values are not positive integers (JS has no concept of long) or positive floats.  Are there any nominal range checks to perform here?

b-ma: > Are there any nominal range checks to perform here?

Yep but if we make sure we go into the rust constructor properly, I think this can be delegated to rust so we don't have to duplicate the values and make the checks twice.

My concern was more about that writing:
```js
const offline = new OfflineAudioContext({ sampleRate: 0, });
```
can't go into the else branch, which is still the case right now (but my comment was quite misleading sorry)

Maybe doing something like:
```js
if (typeof args[0] === 'obj') {
   const options = args[0];
   if ('numberOfChannels' in options && 'length' in options && 'sampleRate' in options) {
      // here we can delegate the error handling to rust (which is still in the todo list...)
      super(numberOfChannel, length, sampleRate);
   } else {
       throw Error('NotSupportedError - Invalid options, must contain blah, blah...');
   }
} else {
   super(...args);
}
```
would work?

b-ma: > JS has no concept of long

Just to note, JS Number are 64 bits, what we actually don't have are floats :)

b-ma: Actually forget my comment, I'm wrong here... `undefined` is not a `positiveInt`

b-ma:
ok let's merge! thank!

PR #17, opened 2023-02-22
Fix the AudioParam method names

ArtBIT:
`cancelScheduledValues` and `cancelAndHoldAtTime` should start with a lowercase letter https://developer.mozilla.org/en-US/docs/Web/API/AudioParam


b-ma:
Nice catch, thanks!

b-ma:
released in 0.7.0

PR #18, opened 2023-03-13
Update napi min required version, 2.0 is not sufficient

orottier:


b-ma:
Damn I forgot that one, thanks!

PR #25, opened 2023-07-17
Use array buffer to be closer to WebAudio API

emptyflash:
Ended up needing this change to work with already existing WebAudio projects. Let me know if it makes more sense to maintain backward compatibility.

https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext/decodeAudioData

### DIFF
--- a/generator/templates/audio_context.tmpl.rs
+++ b/generator/templates/audio_context.tmpl.rs
@@ -1,4 +1,4 @@
-use std::fs::File;
+use std::io::Cursor;
 
 use napi::*;
 use napi_derive::js_function;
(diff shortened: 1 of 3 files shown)


b-ma:
Hey, thanks!

Yes it makes a lot of sens and is actually spec compliant. The current implement was a dirty hack to just make it work and the `load` function was dedicated to disappear anyway.
If I understand well, with your changes we can then rely on a raw `fs.readFileSync` or something? e.g.:

```js
const arrayBuffer = fs.readFileSync('my-file.wav').buffer;
const audioBuffer = await audioContext.decodeAudioData(arrayBuffer);
```

Will just need to update the examples but it won't be a huge issue

b-ma:
Works like a charm, thanks again!

b-ma:
released in v0.11.0

b-ma:
For info, mocking the removed `load` function inside an application is a matter of adding:

```js
import fs from 'node:fs'

function load(filename) {
  return fs.readFileSync(filename).buffer;
}
```

PR #37, opened 2023-10-01
Update upstream crate with `mut` nodes

b-ma:
Hey @orottier 

First attempt to update the bindings regarding the `mut` thing, but it appears to be a bit more tricky than I naively expected...  (I removed most of the nodes generation to keep only the `AudioBufferSourceNode` so it's easier to focus on the problems)

So, I changed the `Rc<...Node>` pattern to `Rc<RefCell<...Node>>` as discussed in https://github.com/orottier/web-audio-api-rs/issues/344, but I fall into a borrow checker traps I don't find how to get rid off.

Main problems are:
- `AudioNode::unwrap()` (and the new `unwrap_mut()`) accessors:
```rs
error[E0515]: cannot return reference to temporary value
  --> src/audio_buffer_source_node.rs:72:9
   |
72 |         &inner.borrow()
   |         ^--------------
   |         ||
   |         |temporary value created here
   |         returns a reference to data owned by the current function
``` 
I think this one is not the biggest issue as these methods are mainly convenience accessors, so in worst case I think I could just get rid of them and replace the code everywhere.

- The `ParamGetter` pattern used to access the parameter is more problematic, same issue but I don't really see how to workaround it:
```rs
error[E0515]: cannot return value referencing temporary value
  --> src/audio_param.rs:22:9
   |
22 | /         match *self {
23 | |             ParamGetter::AudioBufferSourceNodePlaybackRate(ref node) => {
24 | |                 node.borrow().playback_rate()
   | |                 ------------- temporary value created here
25 | |             }
26 | |             ParamGetter::AudioBufferSourceNodeDetune(ref node) => node.borrow().detune(),
27 | |         }
   | |_________^ returns a value referencing data owned by the current function
```

It is a bit frustrating as we don't even need the mutable semantic for `AudioParam`. Maybe a solution would be to replace this pattern with a big macro (just has with the connect / disconnect methods)?

Aside, a perfect solution I think would be that the `AudioNode` wrappers just own the instance, e.g. `NapiAudioBufferSourceNode(AudioBufferSourceNode)`, and the `AudioParam` getter a non mutable reference to the instance, e.g. `AudioBufferSourceNodePlaybackRate(Rc<AudioBufferSourceNode>)`. But if it is possible to do such thing (and I'm really not sure it is), I didn't find any way to implement it... 

If you could have a rapid look, any idea / feedback would be welcome!


orottier:
A solution I would like to check if I could implement `clone` on an AudioParam. This is doable (and we could possibly hide it behind a non-documented feature flag).
In that case, The AudioParamGetter is probably no longer needed. You can just store the plain AudioParam, and not refer to the node
Will investigate more

orottier:
I have prepared a branch `feature/clone-for-audioparam` on the main library so AudioParam can be cloned into an owned value.
I don't know enough about napi to make this work, but I'm convinced if we now stop wrapping the node inside Rc/RefCell, and we store the owned AudioParam alongside in a named property, life should be easier. Maybe you could have a look too?

b-ma:
Cool thanks! I will have a try next week, I let you know

b-ma:
Ok looks like it works! Really nice indeed, it allowed to simplify all the wrappings.
If you are ok with this solution on the upstream side, I'll just wait for the next release to merge this then.

Thanks!

b-ma:
@orottier do you think you could merge this `feature/clone-for-audioparam` branch before next release? 
Would be nice if I could update this in the next weeks and not lag too much behind the upstream crate

orottier:
Ah right, already forgot about it. I will merge it soon

b-ma:
Thanks for the help!

PR #42, opened 2023-11-25
Some work in progress on running the web-platform-tests suite for WebAudio

orottier:
- [x] figure out how to run wpt on our bindings
- [x] add all tests via a git submodule
- [x] add a stub implementation for missing bindings (e.g. OfflineAudioContext.suspend) to avoid some crashes
- [x] avoid all other crashes via the filter list
- [x] get the full suite to run
- [x] collect initial report
- [x] bindings for AudioListener
- [x] implement OfflineAudioContext.suspend
- [ ] bindings for event listeners (_edit_)
- [x] make OfflineAudioContext.start_rendering take &self instead of self
- [ ] check if we can run on all wpt files now
- [ ] collect new report
- [ ] can we drop the wpt fork?
- [ ] start working on panics that occur
- [ ] start working on failed assertions
- [ ] start working on missing functionality


b-ma:
Coo thanks! Nice that you found `https://www.npmjs.com/package/wpt-runner`, porting all these things manually to make them work in Node like I tried was a mess

Will have a look soon, some Node stuff can be made more idiomatic and simplified

orottier:
> Will have a look soon, some Node stuff can be made more idiomatic and simplified

Yes please:)

Meanwhile I'm looking for a method to collect and track the test results.
Also to not pollute the repo too much I will probably replace the wpt files with a git submodule

b-ma:
Ok it's a bit more node friendly now, 
- Moved all the code into './bin/test-harness.mjs`
- You will need to `npm install` again (I added some deps) and then `npm run wpt`

> Meanwhile I'm looking for a method to collect and track the test results.

I sketched something using the programmatic API, seems to be the way to go

> Also to not pollute the repo too much I will probably replace the wpt files with a git submodule

Yup maybe that would be better, I think you didn't move all the files yet isn't it?

orottier:
> I sketched something using the programmatic API, seems to be the way to go

Thanks, definitely beats my `grep .. | uniq -c` approach.

>  think you didn't move all the files yet isn't it?

Indeed, we make the NodeJs runtime crash for many tests and the wpt runner is not properly isolated so the whole suite terminates. I will try to see if I can add some process level isolation to the test runner but since my javascript is terrible I probably won't make it work. Alternative is to put all crashing tests into the `filter` method of wpt-harness.js

b-ma:
> Indeed, we make the NodeJs runtime crash for many tests and the wpt runner is not properly isolated so the whole suite terminates. I will try to see if I can add some process level isolation to the test runner but since my javascript is terrible I probably won't make it work. Alternative is to put all crashing tests into the filter method of wpt-harness.js

Hum I see, if you have one crashing test in mind maybe you could just add it, then I can try to see what we can do?

orottier:
In mobile now, but e.g. analysernode invalid constructor tests (in ctor.html file)

b-ma:
ok thanks will try to have a look later today

b-ma:
Didn't really see how fix that in the test runner, so I fixed the issue in the lib itself (which is anyway what we want I guess :)

I also discriminated all errors of kind `threw "Error" instead of EcmaScript error TypeError.` from "real" errors in the output. So the test still not pass but at least we know this is not a behaviour problem. (I really thought a lot on how to fix this as at first sight it seems like an easy win, but I'm afraid the only solution is to rewrap absolutely everything in JS with try / catch to properly handle exotic error types, doable but so boring...)

All in all, at first sight:
- Good new is that many issues comes from the bindings and not from the rust crate. 
- Sad new is that many issues comes from the bindings and not from the rust crate.

(I really like the fact that the ScriptProcessorNode is deprecated but that some tests rely on it :)

orottier:
Thanks for the additional work. Adding a different result type for "expected error ... but got error .." is very useful.

The wpt runner still crashes a lot for stupid reasons (missing bindings) but I also discovered a real bug, nice.

Maybe the best way forward is:
- add all tests via a git submodule
- add a stub implementation for missing bindings (e.g. OfflineAudioContext.suspend) to avoid some crashes
- avoid all other crashes via the filter list
- get the full suite to run
- collect initial report
- work on fixes, record progress
- present work
- ..
- academic profit :)

orottier:
Slowly making progress, I managed to get into the `the-audio-node-interface` directory

I need these changes in the wpt repo, I will make a fork for that later
```
	deleted:    webaudio/the-audio-api/the-audiocontext-interface/audiocontext-not-fully-active.html
	deleted:    webaudio/the-audio-api/the-audiocontext-interface/audiocontextoptions.html
	deleted:    webaudio/the-audio-api/the-audiocontext-interface/constructor-allowed-to-start.html
	deleted:    webaudio/the-audio-api/the-audiocontext-interface/suspend-with-navigation.html
	deleted:    webaudio/the-audio-api/the-mediaelementaudiosourcenode-interface/cors-check.https.html
	deleted:    webaudio/the-audio-api/the-mediaelementaudiosourcenode-interface/no-cors.https.html

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	webaudio/resources/buffer-loader.js
```

That  new file is created with `cp wpt/webaudio/js/buffer-loader.js wpt/webaudio/resources`.

I will create a tracking issue with all issues that make the runner crash (and hence why I deleted these 6 files)

orottier:
For wpt @ https://github.com/web-platform-tests/wpt/commit/83f318a105a1a9c58a0c7bff65607d753d3ad9ad
node bindings @ https://github.com/ircam-ismm/node-web-audio-api/pull/42/commits/ce309377cd36cf913184e1e61eb2d37d3bff0e4d
rust impl @ https://github.com/orottier/web-audio-api-rs/commit/0dab4c66170a18a5e8786f79c7c8aece50a8ccb5

```
  RESULTS:
  - # pass: 3020
  - # fail: 1293
  - # type error issues: 230
 ```

**Comment by b-ma** [comment]
Cool, nice work!

Few question/remarks:
- About the files you deleted in your wtp fork, it wasn't possible to just filter them in the wpt script?
- For the listener, we agree that the way you did it will break the kind of lazy instantiation we have in the rust side. I do not mean to be fixed right away if this is the case, but maybe we should create a new issue to fix that

Let me know if you need some help on some points and/or when you think this is ready to merge

File: src/offline_audio_context.rs [hunk 1]
```diff
@@ -106,7 +106,10 @@ fn constructor(ctx: CallContext) -> Result<JsUndefined> {
 fn get_current_time(ctx: CallContext) -> Result<JsNumber> {
     let js_this = ctx.this_unchecked::<JsObject>();
     let napi_obj = ctx.env.unwrap::<NapiOfflineAudioContext>(&js_this)?;
-    let obj = napi_obj.unwrap();
+    let obj = match napi_obj.0.as_ref() {
+        Some(v) => v,
+        None => return ctx.env.create_double(0.),
+    };
```
b-ma, line 112: I was a bit surprised about this change at first, but I guess it prevents crashing when we do something like this?

```js
const result = await offlineContext.startRendering();
const now = offlineContext.currentTime
```

Seems that all this `Option` stuff was required to work around:

```rs
pub fn start_rendering_sync(self) -> AudioBuffer
```

Did you think it could be resolved at the rust level to have something like this instead:

```rs
pub fn start_rendering_sync(&self) -> AudioBuffer
```

orottier: Yes definitely, let's change that API. And also implement https://webaudio.github.io/web-audio-api/#dom-offlineaudiocontext-suspend because it is used so much in the wpt.

orottier:
>  About the files you deleted in your wtp fork, it wasn't possible to just filter them in the `wpt-harness.mjs` script?

Yes that would probably be better. My intend is to get rid of the fork soon, but we also have this issue of the missing resource file. I think that needs patching in wpt-runner repo.

>  For the listener, we agree that the way you did it will break the kind of lazy instantiation we have in the rust side? I do not mean this should be fixed right away if this is the case, but maybe we should create a new issue to fix that

Hum right, this has a large performance impact. I could probably use your help here. Is there a way to lazily return the Listener on request? Also let's move the Listener bindings to a separate PR, right?

orottier:
I have put the full TODO list on the top comment. Feel free to add more things.

b-ma:
> Hum right, this has a large performance impact. I could probably use your help here. Is there a way to lazily return the Listener on request? Also let's move the Listener bindings to a separate PR, right?

Ok, I just made a hotfix for that reproducing the lazy behaviour from rust (don't know how to make that cleaner, but it seems to work  :).
Maybe would be cleaner to fix things little by little in seperate PRs yes, or we will end with a huge pile of stuff here. Will try to do that for the listener stuff in the week-end

orottier:
Alright, I removed the old AudioListener binding and replaced with the work in #43 
Then I replaced the suspend/resume stubs with an unresolved instead resolved promise (meaning the `then` closure is never executed, instead of before the rendering starts). This is a bit more pure and shows a few more succeeded assertions.
Then I reinstated 11 of the 21 deleted files in the wpt fork because they now don't crash the runner.

```
  RESULTS:
  - # pass: 3179 (+159)
  - # fail: 1341 (+48)
  - # type error issues: 235 (+5)

files ignored: 10 (-11)
```

Previous results: https://github.com/ircam-ismm/node-web-audio-api/pull/42#issuecomment-1834123195

orottier:
@b-ma I changed start_rendering_sync over at https://github.com/orottier/web-audio-api-rs/pull/405
Feel free to take my hotfix of https://github.com/ircam-ismm/node-web-audio-api/pull/42/commits/e57fd25a21aee511e92d307e97eb44ac13c9d9d2 and shape it properly in a PR

```
  RESULTS:
  - # pass: 3200 (+21)
  - # fail: 1347 (+6)
  - # type error issues: 235 (=)

files ignored: 9 (-1)
```

Previous results https://github.com/ircam-ismm/node-web-audio-api/pull/42#issuecomment-1837234986

b-ma:
> @b-ma I changed start_rendering_sync over at https://github.com/orottier/web-audio-api-rs/pull/405
> Feel free to take my hotfix of https://github.com/ircam-ismm/node-web-audio-api/commit/e57fd25a21aee511e92d307e97eb44ac13c9d9d2 and shape it properly in a PR

Ok I'll have a look soon

I merged AudioListener stuff from #43 into main too (is now implemented as proper JsClass rather than JsObject)

b-ma:
Fix for OfflineContext is here: #44

orottier:
New results with webaudio-rs v0.38 en new templating:

```
  RESULTS:
  - # pass: 3285 (+85)
  - # fail: 1444 (+97)
  - # type error issues: 235 (=)

files ignored: 9 (=)
```

Previous results https://github.com/ircam-ismm/node-web-audio-api/pull/42#issuecomment-1837242822

orottier:
New results with suspend/resume and AudioListener.setPosition

```
  RESULTS:
  - # pass: 3386 (+101)
  - # fail: 1380 (-64)
  - # type error issues: 235 (=)

files ignored: 8 (-1)
```

Previous results https://github.com/ircam-ismm/node-web-audio-api/pull/42#issuecomment-1842260849

src/audio_listener.rs
```diff
@@ -112,3 +122,54 @@ fn constructor(ctx: CallContext) -> Result<JsUndefined> {
...
     ctx.env.get_undefined()
 }
+
+#[js_function(3)]
+fn set_position(ctx: CallContext) -> Result<JsUndefined> {
+    // TODO https://webaudio.github.io/web-audio-api/#dom-audiolistener-setposition
+    //
+    // When any of the positionX, positionY, and positionZ AudioParams for this AudioListener have
+    // an automation curve set using setValueCurveAtTime() at the time this method is called, a
+    // NotSupportedError MUST be thrown.
```
b-ma, line 132: I think this is actually ok as the error will be thrown by the `set_value` call. The check is done in the `AudioParam`.

But this is also something that should be fixed on the rust side at some point because the `panic` would occur in the render thread which is not good...

b-ma:
Hey, really good job thanks! It's very nice to have such move forward on that point.

I will just add a small fix on `AudioParam` tomorrow before merging. Then I will refactor the JS code to make it more clean as described in #53, I think it will help to continue on that path more simply

b-ma:
Ok to merge on my side! Good for you?

orottier:
All good 👍

orottier:
New results with all audioparam attributes and OfflineAudioContext::state

```
  RESULTS:
  - # pass: 3413 (+27)
  - # fail: 1353 (-27)
  - # type error issues: 235 (=)

files ignored: 8 (-1)
```

Previous results https://github.com/ircam-ismm/node-web-audio-api/pull/42#issuecomment-1870096858

PR #44, opened 2023-12-03
fix: remove Option from OfflineAudioContext`

b-ma:


orottier:
Looking good, can be published with the rust v0.38 version

PR #49, opened 2023-12-06
Fix/clippy

b-ma:


orottier:
Nice! Can we add a check in CI?

b-ma:
Just added clippy check in ci, cf. https://github.com/ircam-ismm/node-web-audio-api/actions/runs/7156606098

PR #52, opened 2023-12-10
Feature - context `onstatechange` event and listeners

b-ma:
First working version

## Todos

- [ ] create proper `Event` instance (missing `target`, etc)
- [ ] clear listener when the rust instance is dropped? (is it dropped actually?)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,8 +12,8 @@ crate-type = ["cdylib"]
 [dependencies]
 napi = {version="2.13", features=["napi6"]}
 napi-derive = "2.13"
-web-audio-api = "0.38"
-# web-audio-api = { path = "../web-audio-api-rs" }
+# web-audio-api = "0.38"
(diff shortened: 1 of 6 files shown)


orottier:
I was wondering if it may not be easier and more general to perform actual event dispatching, like so:

- when an audiobuffersourcenode is created
- always register an `on_ended` on the rust side
- in on_ended, call `dispatchEvent` on the Js object
- monkey patch the setter of `onended` to an `addEventListener` call

In any case, the current implementation also looks good to me

b-ma:
Hum, that's more or less what is done, but maybe your solution would be more elegant yes, I will have a try

In any case, the current implementation is not clean as it prevents the event target to be garbage collected...

I will also have another look on that article https://www.nearform.com/blog/node-js-and-the-struggles-of-being-an-eventtarget/ which might also help having something more clean

PR #54, opened 2023-12-22
Improve error handling

b-ma:
Some promising tests regarding #8


b-ma:
Hum does not work as expected with method, fall into "illegal invocation" trap... cf. https://mtsknn.fi/blog/illegal-invocations-in-js/

Will need to generate the API on JS side to try / catch native method invocations, e.g.

```js
class AudioBufferSourceNode {
  // ...
  start(when) {
    try {
      super.start(when);
    } catch (err) {
      throwSanitizedError(err);
    }
  }
}
```

Some new tests there: https://github.com/ircam-ismm/node-web-audio-api/tree/feat/error-handling

Issue #57, opened 2023-12-28
The web-platform-tests tracking issue


orottier:
The [wpt suite](https://github.com/web-platform-tests/wpt) was introduced to this lib in #42. It runs without crashing now and our aim is to improve the test score over time.

This tracking issue lists the high level TODOs. Single test failures are probably best reported and fixed with separate issues (either in the node bindings or the rust library).

- [ ] get rid of the wpt fork (how to deal with `webaudio/resources/buffer-loader.js`, reinstate 8 deleted files)
- [ ] how to handle the `crashtest` tests?
- [ ] collect results of the browser implementations to determine baseline performance
- [ ] implement major missing functionality (event handlers, worklets, ..)
- [ ] improve reported errors (see #8 #54 )
- [ ] check reporter, it seems to count some log lines as errors e.g. `× Executing "test"`
- [ ] focus on foundational test failures (`the-audionode-interface`, `the-audioparam-interface`, `the-audiocontext-interface`, `the-offlineaudiocontext-interface`, `processing-model`, `the-audiobuffer-interface`)
- [ ] focus on other test failures
- [ ] run WPT in CI?

b-ma:
Thanks for the follow up, two notes:

>  get rid of the wpt fork (how to deal with webaudio/resources/buffer-loader.js, reinstate 8 deleted files)

I think this can be done by 
1. Implementing the `context.decodeAudioData(response, successCallback, errorCallback)` signature, which should be rather simple to do on the JS side
2. Providing a mock to `XMLHttpRequest` to the runner using raw `fs.readFileSync` internally

>  collect results of the browser implementations to determine baseline performance

Maybe a bit old but there is this report: https://webaudio.github.io/web-audio-api/implementation-report.html

orottier:
I have submitted a PR over at wpt for the buffer-loader.js: https://github.com/web-platform-tests/wpt/pull/43824
Let's see where that gets us

b-ma:
results from #42 

> RESULTS:
>  - # pass: 3413 (+27)
>  - # fail: 1353 (-27)
>  - # type error issues: 235 (=)
>
> files ignored: 8 (=)

b-ma:
results with #61 

```
  RESULTS:
  - # pass: 3605 (+192)
  - # fail: 1284 (-69)
  - # type error issues: 234 (-1)
```

b-ma:
results with #66 
```
RESULTS:
  - # pass: 3778 (+173)
[...]

PR #60, opened 2023-12-28
feat: add legacy decodeAudioData signature

b-ma:
First step regarding #57 

> "get rid of the wpt fork (how to deal with webaudio/resources/buffer-loader.js)"

### DIFF
--- a/.scripts/wpt-mock/XMLHttpRequest.js
+++ b/.scripts/wpt-mock/XMLHttpRequest.js
@@ -1,0 +1,39 @@
+const fs = require('node:fs');
+// to be passed to wtp-runner step
+// window.XMLHttpRequest = XMLHttpRequest;
+class XMLHttpRequest {
+  constructor() {
+    this._pathname;
+    this._onload;
+    this._onerror;
+    this.response;
+  }
+
+  open(_protocol, url) {
+    this._pathname = url;
+  }
+
+  send() {
+    let buffer;
+
+    try {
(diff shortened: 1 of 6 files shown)


b-ma:
ok, I just added a mock for the XMLHttpRequest in `.scripts/wpt-mock/`, should allow to fix the second step

orottier:
Hah pretty cool. To be fair I had not noticed yet that we need this machinery.

The issue I was referring to was the problem that we need a `cp wpt/webaudio/js/buffer-loader.js wpt/webaudio/resources` in order for the wpt-runner to not crash. That is still something to be addressed (probably by rebuilding the statis assets of the wpt-runner repo - I need to look into this further).

The cool thing is that you now have probably solved some failed tests because of the missing XMLHttpRequest :)

b-ma:
Hum weird indeed... This looks like an issue in the `wpt` repo indeed, the path in another file is right https://github.com/web-platform-tests/wpt/blob/0c8c0662ed5d444bf12992e433502e8b6af640d4/webaudio/the-audio-api/the-mediaelementaudiosourcenode-interface/mediaElementAudioSourceToScriptProcessorTest.html#L20

And in the file you pointed into the `BufferLoader` is not even used...

As a temporary solution, we could just copy the file in the `resources` directory using npm scripts (should be very simple) and ignore changes in the `wpt` submodule (e.g. https://medium.com/@maryam.ghani160/how-to-ignore-changes-in-git-submodules-a3ea1463bdde), what do you think?

PR #61, opened 2023-12-29
Implement EventTarget interface

b-ma:
Just another try compared to #52 

Tried to follow your idea, seems a bit better to me indeed but there are still some convoluted things I didn't find how to work around (no big deal though, it works... :)

_~Note that it depends on [415](https://github.com/orottier/web-audio-api-rs/pull/415)~ (merged)_

List of all events in the spec:

- [ ] BaseAudioContext::statechange
    - [x] AudioContext::statechange
    - [ ] OfflineAudioContext::statechange (requires https://github.com/orottier/web-audio-api-rs/issues/416)
- [x] AudioContext::sinkchange
- [ ] OfflineAudioContext::complete (requires https://github.com/orottier/web-audio-api-rs/issues/411)
- [x] AudioScheduledSourceNode::ended

todo later

- _AudioRenderCapacity::update (not wrapped yet)_
- _AudioWorkletNode::processorerror (not wrapped yet)_
- _ScriptProcessorNode::audioprocess (not implemented yet)_


b-ma:
One thing that I found that is more a concern on the rust side, is that the first "resume" `statechange` event is sometimes triggered, sometimes not. I think this some kind of race condition but this is a bit weird from a user perspective.

I will try to make a failing test for that later

src/audio_context.rs
```diff
@@ -602,3 +613,39 @@ fn create_media_stream_source(ctx: CallContext) -> Result<JsObject> {
...
+
+    let context_clone = context.clone();
+    context.set_onstatechange(move |e| {
+        tsfn.call(Ok(e), ThreadsafeFunctionCallMode::Blocking);
+
+        if context_clone.state() == AudioContextState::Closed {
+            // We need to clean things around so that the js object can be garbage collected.
+            // But we also need to wait so that the previous tsfn.call is executed,
+            // this is not clean, but don't see how to implement that properly right now.
+            std::thread::sleep(std::time::Duration::from_millis(100));
```
orottier, line 644: I hope this hack is only needed for the onStateChange event? Or also for other types?

orottier (approved):
This all looks quite complicated but I guess this is due to the fact that NodeJs has no JS-alike event targets?

orottier:
> One thing that I found that is more a concern on the rust side, is that the first "resume" statechange event is sometimes triggered, sometimes not. I think this some kind of race condition but this is a bit weird from a user perspective.

I could make a start with https://github.com/orottier/web-audio-api-rs/issues/410 where the whole thing will be rewritten anyway

b-ma:
> This all looks quite complicated but I guess this is due to the fact that NodeJs has no JS-alike event targets?

Actually, this is quite boring because there is a compliant `EventTarget` implementation in Node, but then we would need multiple inheritance to inherit from both `EventTarget` and the NAPI classes, which is not possible in JS. 

A solution is described [here](https://www.nearform.com/blog/node-js-and-the-struggles-of-being-an-eventtarget/), but this is something internal to Node and the given "trick" is not exposed... so we have work around somehow... Maybe I could try to open an issue on the Node repo to see if there is a solution I didn't find

src/audio_context.rs
b-ma: Hum no it seems to be required for other types as well, need to investigate further

b-ma:
I don't really understand why I have less tests working than you... but that was already the case before

```
  RESULTS:
  - # pass: 3304 (-107)
  - # fail: 1251 (-103)
  - # type error issues: 214 (-21)
```

Could you run the wpt see what you get?

In any case, there is already a lot (too much) done in this PR, I will merge once https://github.com/orottier/web-audio-api-rs/pull/415 is released if ok for you

orottier:
> I don't really understand why I have less tests working than you... but that was already the case before

The wpt submodule is out of date on your end I think, and this was accidentally merged into main over at fd4d71085714700946e1f0a5e8ecddfe42d134a2

I will push the correct version and I think you should then run `git submodule update --remote --merge`

I now notice we have a fatal wpt crash at 
```
node:internal/process/promises:289
            triggerUncaughtException(err, true /* fromPromise */);
            ^
Error {
  message: 'assert_equals: state on suspend fulfilled. expected "suspended" but got "running"',
  stack: '    at http://127.0.0.1:53923/webaudio/the-audio-api/the-audiocontext-interface/suspend-after-construct.html:53:5\n' +
    '    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)'
}
```

So I don't have the results right now, will look into it later today

orottier:
> In any case, there is already a lot (too much) done in this PR, I will merge once [orottier/web-audio-api-rs#415](https://github.com/orottier/web-audio-api-rs/pull/415) is released if ok for you

I have just released v0.40.0 for you

b-ma:
Thanks!

Seems that it was the wpt that was not up to date indeed. 

For the error you encounter, I don't really understand this works well on my side. But I have seen yesterday that building in debug mode could create weird errors and crashes with wpt, I updated the wpt script (`npm run wpt`) so that it build in release mode, a bit longer but no strange issues of that kind, and we can always run `npm run wpt:only` to bypass the build step if needed.

Updated results on my side are:

```
  RESULTS:
  - # pass: 3605 (+192)
  - # fail: 1284 (-69)
  - # type error issues: 234 (-1)
```

PR #70, opened 2024-01-12
Drop wpt fork

orottier:
https://github.com/web-platform-tests/wpt/pull/43824 was merged and somehow we don't crash the runner for other 404s.
I have listed three remaining files to ignore in the wpt-harness.js now. I will add those as todo to #57 

```
  RESULTS:
  - # pass: 3856 (-2) 
  - # fail: 1226 (+10)
  - # type error issues: 63 (+1)
  - # files ignored: 3 (-5)
```

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "wpt"]
 	path = wpt
-	url = git@github.com:orottier/wpt.git
+	url = git@github.com:web-platform-tests/wpt.git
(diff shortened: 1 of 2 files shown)


orottier:
The drop in `# pass` is due to `wpt/webaudio/the-audio-api/the-waveshapernode-interface/waveshaper.html` not running, it also 404s on the buffer loader. I will try to get that one, and the other 404s fixed in the upstream repo as well. This can be merged in the meantime

b-ma:
Cool, thanks!

> somehow we don't crash the runner for other 404s.

Actually I installed in the XHR mock in some previous commit, maybe it solved some issues
https://github.com/ircam-ismm/node-web-audio-api/blob/3837a264fc2764ce780700b4a92721947a2d1413/.scripts/wpt-harness.mjs#L50

PR #76, opened 2024-02-05
Upgrade to 0.42, fix wpt checkout, add channelCount to MediaTrackConstraints

### LINKED ISSUE #74: channelCount constraint of getUserMedia is ignored
When using an audio interface with more than 2 channels, getUserMedia should be able to return a media stream with more than 2 channels by providing a channelCount constraint. Browsers generally don't support more than 2, but I was hoping since the spec does support it that this project might achieve it.

Here's some test code. One of my audio devices has 18 inputs, but it reports 2. I tired using `18`, `{ exact: 18 }` and `{ ideal: 18 }` as constraints, but all were ignored.

```TypeScript
import { mediaDevices, AudioContext } from 'node-web-audio-api';

async function init() {

	const devices = await mediaDevices.enumerateDevices();

	console.log( audioContext.destination.channelCount, 'outputs' );

	for ( const device of devices ) {
		if ( device.kind === 'audioinput' ) {
			console.log( device.label );
			const mediaStream = await mediaDevices.getUserMedia( {
				audio: {
					deviceId: device.deviceId,
					channelCount: 18
				}
			} );
			const source = audioContext.createMediaStreamSource( mediaStream );
			console.log( '>', source.channelCount, 'inputs' );
			source.disconnect();
		}
	}

	audioContext.close();
}

init()
	.then( () => process.exit() )
	.catch( console.error );

```

orottier:
@trevorparscal could you have a look if this works for you? I don't have a multichannel mic at hand here and the underlying `cpal` library is surprisingly relaxed in allowing exotic channel counts - so I can't really test properly myself.

Fixes #74

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ version = "0.15.0"
 crate-type = ["cdylib"]
 
 [dependencies]
-napi = {version="2.14", features=["napi9", "tokio_rt"]}
+napi = {version="=2.14", features=["napi9", "tokio_rt"]}
 napi-derive = "2.14"
 uuid = {version="1.6.1", features = ["v4","fast-rng"]}
 web-audio-api = "0.42"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@ crate-type = ["cdylib"]
 napi = {version="2.14", features=["napi9", "tokio_rt"]}
 napi-derive = "2.14"
 uuid = {version="1.6.1", features = ["v4","fast-rng"]}
-web-audio-api = "0.41.1"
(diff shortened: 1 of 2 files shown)


orottier:
Ready for merging!

b-ma:
Cool thanks!

Will try to set up some example with Pure Data or Max patch
And will create an issue regarding [1fd1e22](https://github.com/ircam-ismm/node-web-audio-api/pull/76/commits/1fd1e2277f0ebebeb7ebfc0d9eb3a012fe0d87b7), if I make it `AudioBuffer::getChannelData` safe and return a non mutable `Float32Array` this breaks lots (~500...) of wpt test

PR #77, opened 2024-02-05
Revert "chore(deps-dev): bump octokit from 2.0.14 to 3.1.2"

orottier:
Reverts ircam-ismm/node-web-audio-api#56


b-ma:
Oh damn, rather boring as it is not used in the CI, anyway thanks for the catch!

Issue #78, opened 2024-02-06
Fix AudioBuffer::getChannelData with napi-rs 2.15


b-ma:
cf. hotfix [1fd1e22](https://github.com/ircam-ismm/node-web-audio-api/pull/76/commits/1fd1e2277f0ebebeb7ebfc0d9eb3a012fe0d87b7)

The underlying `napi-rs` change seems to be this one [ac3626a](https://github.com/napi-rs/napi-rs/commit/ac3626a023fe197f3ab18152aede2372542d5c0b) where it seems that we now need to have a mutable handle to the values, ~probably requires some change in upstream crate too.~ cf. `AudioBuffer::get_channel_data_mut`

The alternative of assuming that `getChannelData` returns an immutable handle to the channel buffer break many wpt tests

b-ma:
Fixed with #79

PR #79, opened 2024-02-06
fix: getChannelData - cf. #78

b-ma:


orottier:
I think this inhibits some undefined behaviour - theoretically.

You are asking the compiler for a mutable reference of the AudioBuffer, but then this information is 'erased' via the unsafe slice cast. This means if you call the method twice and mutate one of the copies, the compiler could assume the other copy will not be affected.

I don't have a clear solution though (perhaps return the same TypedArray via `set_property`??) - should we add an issue to address this later?

b-ma:
Yup I really agree this is a bit a hacky solution, but I can't think of any other one right now

At least for now it just behaves as it was before, we can update napi-rs and we don't loose wpt tests (that quite heavily rely on this behavior...)

Probably adding an issue to keep this in mind is a good idea yup

PR #82, opened 2024-02-21
Implement AudioNode interface for AudioDestinationNode

### LINKED ISSUE #81: DestinationNode needs js `AudioNodeMixin` to handle panics
From #57 the tests in `the-destinationnode-interface` fail because the `audioContext.destination.channelCount = 0;` throws a rust panic that is not wrapped by a JS error. I looked in the `js/` directory and there is no trace of DestinationNode. Should it be added?

b-ma:
Fix #81 

Also fix another issue with the Errors

```
> before
RESULTS:
- # pass: 4276
- # fail: 761
- # type error issues: 57
  
> after
RESULTS:
  - # pass: 4282
  - # fail: 755
  - # type error issues: 57
 ```

### DIFF
--- a/.scripts/wpt-harness.mjs
+++ b/.scripts/wpt-harness.mjs
@@ -7,10 +7,7 @@ import * as nodeWebAudioAPI from '../index.mjs';
 
 // mocks
 import createXMLHttpRequest from './wpt-mock/XMLHttpRequest.js';
-// import {
-//   NotSupportedError,
-//   InvalidStateError,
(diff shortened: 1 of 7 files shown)


b-ma:
The fix is not as clean as I would like (duplicated code) but don't really see how to fix that right now

orottier (approved):
Hmm okay, the code duplication is indeed smelly but perhaps we write an issue to look into it later?

b-ma:
Yup

PR #88, opened 2024-02-24
Fix periodic wave

b-ma:
see `the-periodicwave-interface`
- before : 23 / 33 successful tests
- after : 31 / 33 successful tests


orottier (approved):
Monkey sees wpt improvement, monkey approves

Issue #84, opened 2024-02-21
Consider moving argument type check to JS side


b-ma:
- Easier to implement in JS compared to Rust
- Should simplify the Rust code to a great extent
- Handle type coercion in a more idiomatic way (i.e. setValueCurveAtTime should support Array, Float32Array, Float64Array, etc.)

See `AudioNode` constructor `option` argument as an example (cf. 4b2dc95)

- [x] Handle audio node `context` argument
- [x] Handle audio nodes options argument
- [ ] Handle `AudioNodeOption`

b-ma:
cf. [23a3b6e](https://github.com/ircam-ismm/node-web-audio-api/pull/90/commits/23a3b6e0287eb1537ab4c2d5c47f1a2d112e1c2c) for cleaner pattern in rust side

PR #90, opened 2024-02-26
Feat: wrap AudioBuffer with JS proxy

b-ma:
### Before

```
RESULTS:
  - # pass: 115
  - # fail: 23
  - # type error issues: 20
 ```
### After 
(w/ https://github.com/orottier/web-audio-api-rs/pull/463) 

```
  RESULTS:
  - # pass: 145
  - # fail: 13
  - # type error issues: 0
```

### CODE REVIEW

File: generator/js/BaseAudioContext.mixin.tmpl.js [hunk 1]
```diff
@@ -40,6 +41,10 @@ ${d.nodes.map(n => `    ${d.name(n)},`).join('\n')}
       }
     }
 
+    createBuffer(numberOfChannels, length, sampleRate) {
```
orottier, line 44: Good to have this constructor option supported

js/lib/cast.js
```diff
@@ -0,0 +1,20 @@
+
+exports.toSanitizedSequence = function toSanitizedSequence(data, targetCtor) {
```
orottier, line 2: What is this for?

b-ma: Finally... for nothing in this PR :) but I was thinking `copyFromChannel` and `copyToChannel` had the same API as for example https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-createiirfilter which support any kind of floating point value sequence, e.g. Array, Float32Array, Float64Array (numbers is JS are basically only `double`)
Should be useful later I imagine

b-ma:
Just realized there is an issue with `startRendering` and `decodeAudioData` which do not return a wrapped object.
Need to change the wrapping approach to make it more like a mixin rather than use inheritance I guess (e.g. like 
 `AudioParam.js`)

orottier:
Without looking into it, I just want to comment it is possible to both decrease in success and fail.
One newly introduced failure can block many subsequent successes.

I'm using the following trick to digest the results better:

(make sure to run `npm run build` before)

create file `run-wpt.sh` and `chmod +x run-wpt.sh`:
```
#!/bin/bash
echo $1
echo ~~~~~~~~~~~~~~~~~
node ./.scripts/wpt-harness.mjs --filter "$1" 2>&1 | grep -A 3 RESULTS
echo
```

dissect the full suite per directory:
`ls wpt/webaudio/the-audio-api | xargs -I {} ./run-wpt.sh {}`
check the specific file results in a dir:
`ls wpt/webaudio/the-audio-api/the-oscillatornode-interface | xargs -I {} ./run-wpt.sh {}`

b-ma:
Ah thanks!

b-ma:
Ok, finally got it working, it was a bit more complex than expected but paves the way for #84

#### main

```
RESULTS:
  - # pass: 5215
  - # fail: 628
  - # type error issues: 68
```

#### feat/Audiobuffer-proxy

```
RESULTS:
  - # pass: 5262
  - # fail: 603
  - # type error issues: 46
```

orottier:
Cool!

PR #94, opened 2024-03-12
Fix: media stream source node

b-ma:
Was broken as it was not generated and wrapped just as other nodes

(no impact on wpt)


orottier (approved):
I don't grok all the changes but having it working is nice :)

b-ma:
huhu, yup that's a mess... really need to clean all this a bit at some point

PR #95, opened 2024-03-13
Feat: ended events for AudioSheduledSourceNode in OfflineAudioContext

b-ma:
Maybe not to be merged right now as it is not tested. I was thinking that the problem was solved in upstream crate, but it seems this is still not the case, could you confirm or did I miss something?

This issue triggers some timeout in the wpt, e.g. `wpt/webaudio/the-audio-api/the-audiobuffersourcenode-interface/audiobuffersource-ended.html`


orottier:
Hey, the OfflineAudioContext only spawns statechange/completed events (that was what I recently added).
It does not handle AudioNode events to keep things efficient/simple - we could add that later but I don't see a use case for it

orottier:
The problem being: we either need to spawn a new thread per OfflineAudioContext for the event loop - which is quite a bit of overhead. Or we don't use a thread, but buffer all the events and run the handlers right before `oncomplete` but then it does not make a lot of sense for the end user.

b-ma:
> we could add that later but I don't see a use case for it

Yup I agree, the use case is rather hard to imagine... This is mostly for wpt, the timeout for each `ended` event test is around ~10s which is rather boring

If we go to resolve this, I would vote for spawning a thread so we can just share the logic with the online AudioContext (if it is possible)

I will just convert this to draft until we decide something

orottier:
This PR can be tried again now, the offline event loop is on `main` in the base lib

b-ma:
Cool, I will have a new look soon

b-ma:
Seems to be some improvements

#### main
```
RESULTS:
  - # pass: 6848
  - # fail: 706
  - # type error issues: 5
> wpt duration: 2:22.697 (m:ss.mmm)
```

#### feat/ended-events
```
RESULTS:
  - # pass: 6854
  - # fail: 704
  - # type error issues: 5
> wpt duration: 2:08.718 (m:ss.mmm)
```

(both are built against rs upstream main)

But there are still some issues. In short we have two cases we don't handle correctly (just take an `OscillatorNode` for the example, but this is the same behaviour for all `AudioScheduledSourceNode`s)

1. Explicit stop before the end of rendering works

```js
const context = new OfflineAudioContext(1, 48000, 48000);
const src = new ConstantSourceNode(context);
src.connect(context.destination);
src.onended = () => console.log('onended fired 1');

src.start();
src.stop(0.9);

context.startRendering();
```

2. Explicit stop at end of the rendering fails

```js
const context = new OfflineAudioContext(1, 48000, 48000);
const src = new ConstantSourceNode(context);
src.connect(context.destination);
src.onended = () => console.log('onended fired 1');

src.start();
src.stop(1.);

context.startRendering();
```

This one is rather logical, as we fire the ended event on next render quantum, which in this case is never computed. We could fix it either by reviewing when the `ended` events are send, or by doing a last dry run at the end of the offline context (a bit dirty I agree :)

3. Implicit stop fails

```js
const context = new OfflineAudioContext(1, 48000, 48000);
const src = new ConstantSourceNode(context);
src.connect(context.destination);
src.onended = () => console.log('onended fired 1');

src.start();
// no explicit stop

context.startRendering();
```

Not sure how to handle that case (is present in wpt, cf. the-audiobuffersourcenode-interface/audiobuffersource-ended.html), maybe have the offline duration in render scope and use it to clamp stop time?

In any case, on this side the logic seems to work so I think I will merge, we can iterate on the failing cases later!

orottier:
Cool. Yeah for those two cases we can create an issue at the base lib!

b-ma:
@todo - register `AudioScheduledSourceNode` listener only on start

orottier:
I have created https://github.com/orottier/web-audio-api-rs/issues/489

b-ma:
> I have created https://github.com/orottier/web-audio-api-rs/issues/489

+1

orottier (approved):
LGTM. I'm working on some graph stuff in the meantime but if you prefer a release earlier just let me know

b-ma:
> I'm working on some graph stuff in the meantime but if you prefer a release earlier just let me know

That would be really nice yes, I could merge all things based on that (especially #107 which is quite a huge one)

orottier:
web-audio-api v1.0.0-rc.4 is available

b-ma:
Cool thanks!

PR #99, opened 2024-03-25
Feat - Implement proper inheritance chain

### LINKED ISSUE #97: AudioNode class is not exported and not shared between node classes
The AudioNode base class is not exported. The WebAudio spec exposes it.

This makes it e.g. tricky to check if an object is an instance of AudioNode and doesn't support [some hacks for composing nodes](https://github.com/GoogleChromeLabs/web-audio-samples/wiki/CompositeAudioNode).

This is probably because each node class a different parent class (all named AudioNode though). There's probably a good reason to do this, but it does break the spec.

```javascript
import { AudioContext, OscillatorNode, GainNode } from 'node-web-audio-api';

// Get GainNode superclass to try to get AudioNode
const AudioNodeHack = Object.getPrototypeOf(GainNode);

const gain = new GainNode(audioContext);
const oscillator = new OscillatorNode(audioContext);

// A GainNode is of course an instance of its parent class...
console.log(gain instanceof AudioNodeHack); // true

// ...but an OscillatorNode is not an instance of GainNode's parent
// because a different AudioNode parent class is created for
// each node class
// https://github.com/ircam-ismm/node-web-audio-api/blob/f6fcdd5099af889103ad4e59e929f2bba7c39590/js/AudioNode.mixin.js#L34
console.log(oscillator instanceof AudioNodeHack); // false
console.log(
    Object.getPrototypeOf(GainNode) === Object.getPrototypeOf(OscillatorNode)
    ); // false

audioContext.close();
```

b-ma:
Review the architecture to use facades instead of mixins to better follow spec inheritance (should fix #97)

#### main
```
RESULTS:
  - # pass: 5406
  - # fail: 569
  - # type error issues: 33
```

#### feat/inheritance-chain
```
RESULTS:
  - # pass: 5415
  - # fail: 565
  - # type error issues: 30
```

@jampekka could you confirm it works for your use-case? If it does having a simple example would be really welcome!


js/EventTarget.js
```diff
@@ -1,23 +1,17 @@
 const { isFunction } = require('./lib/utils.js');
+const { kNapiObj } = require('./lib/symbols.js');
 
 const kEventListeners = Symbol('node-web-audio-api:event-listeners');
 const kDispatchEvent = Symbol.for('node-web-audio-api:napi-dispatch-event');
 
-module.exports = (superclass, eventTypes = []) => class EventTarget extends superclass {
+class EventTarget {
```
jampekka, line 7: Would it be possible to use the native EventTarget with the new wrapping? This would make it even more spec compliant.

jampekka:
Whoa, that was fast! Per quick look seems very nice. I'm very busy today, OK if I review and test tomorrow?

js/EventTarget.js
b-ma: That would make sens indeed, but it requires some work / testing to bind that properly with the native Rust events. I will open a separate issue

b-ma:
No problem, let me know!

tests/junk.mjs
orottier, line 1: We could rewrite this wonderful file to a test suite once maybe :)

js/EventTarget.js
b-ma: Done in #[3bf361e](https://github.com/ircam-ismm/node-web-audio-api/pull/99/commits/3bf361e861843f97f77debb27b41ecd6885c25d7)

tests/junk.mjs
b-ma: Huhu :) ok I should probably ignore it...

orottier (approved):
Again, I have not studied all the details but the overall idea and the wpt results look good to me!

jampekka:
Can confirm that checking for `AudioNode` parent works for my case. However, `AudioNode` is not exported, so I have to dig it out by `const AudioNode = Object.getPrototypeOf(GainNode)`. I'll try if I can manage to fix this.

jampekka:
Works with `AudioNode.connect` monkeypatching too. See #101 and https://github.com/jampekka/node-web-audio-api/blob/199c5ee5556f3b09a011c4d8f821fe6cabc4aa0b/examples/composite-audio-node.mjs

Not sure if this should be an example though. Perhaps rather a test? Regarding those, `npm test` doesn't work currently, changing the test script to `mocha tests` does fix this.

P.S. Is making a pull request against this pull request branch the proper way to suggest code changes in review?

b-ma:
> Not sure if this should be an example though.

I think it's nice to have as an example, I will have with another approach which I think is more clean (i.e. extending GainNode or seomthing), but as it's promoted by google dev its nice to have here

> Regarding those, npm test doesn't work currently, changing the test script to mocha tests does fix this.

Yup, I always run them explicitely `npm test -- tests/*.spec.js`, maybe should be added in README ?

b-ma:
All flags green :) Thanks

jampekka:
> I think it's nice to have as an example, I will have a test with another approach which I think could be a bit more clean (i.e. extending GainNode or something), but as it's proposed by google dev its nice to have here anyway

Extending GainNode should work. But for composition I think you still have to hack the .connect somehow. The Google dev proposal is from 2016, when the spec (or implementations) maybe didn't support AudioNode constructors, so there may be cleaner ways now, although I didn't find one.

> Yup, I always run them explicitely npm test -- tests/*.spec.js, maybe should be added in README ?

Would be helpful for new contributors. What's the reason not to have this as the npm test script? Does that e.g. get automatically run by some toolchain but we don't want to run them?

(I think you mean pm test -- tests/*.spec.**m**js)

b-ma:
> Extending GainNode should work. But for composition I think you still have to hack the .connect somehow. The Google dev proposal is from 2016, when the spec (or implementations) maybe didn't support AudioNode constructors, so there may be cleaner ways now, although I didn't find one.

I've never really tried myself, but that's something I'm quite interested in for creating audio libs, that's the occasion :)

> Would be helpful for new contributors. What's the reason not to have this as the npm test script? Does that e.g. get automatically run by some toolchain but we don't want to run them?

Don't really know, probably just that I'm used to it... let's do that, this is more simple, we can always call mocha directly if needed

PR #101, opened 2024-03-30
Expose AudioNode class

jampekka:
Exposing AudioNode class to JS so we can do `node instanceof AudioNode`. Did not test the monkey patching yet.


b-ma:
Cool thanks!

PR #103, opened 2024-04-01
Feat: run idlharness from wpt

### LINKED ISSUE #86: How to run the idl-harness from the wpt suite?
The file `wpt/webaudio/idlharness.https.window.js` contains 1166 checks and we are not able to run them via the wpt-runner, it seems. This skews the test score comparison to the browser implementations

b-ma:
Fix #86

```
$ npm run wpt:only -- --filter idlharness.https.window.html

RESULTS:
  - # pass: 613
  - # fail: 530
  - # type error issues: 0
```

Was finally just a missing mock for `fetch`, but a bit tricky to find as the `idlharness.js` file is copied inside wpt-runner 

(got lucky to find it, I think have stumbled on that "not fiddling in the right file and getting mad" quite a few times already...)

### DIFF
--- a/.scripts/wpt-harness.mjs
+++ b/.scripts/wpt-harness.mjs
@@ -7,6 +7,7 @@ import * as nodeWebAudioAPI from '../index.mjs';
 
 // mocks
 import createXMLHttpRequest from './wpt-mock/XMLHttpRequest.js';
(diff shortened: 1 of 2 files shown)


orottier (approved):
Hooray!

Issue #91, opened 2024-03-04
MediaStream Destination Node


justin0mcateer:
You mention the MediaStreams API is somewhat out of scope for this library, but there is an implementation (wrapper) of the MediaStream Source Node. Also, the upstream/wrapped crate appears to implement the MediaStream Destination node:
https://github.com/orottier/web-audio-api-rs/blob/main/src/node/media_stream_destination.rs

Aside from perhaps needing to be written manually (vs generated) what are the challenges with creating a MediaStream Desitnation Node wrapper?  Is there anyone able to take this on if it were to be sponsored?

b-ma:
Hey, as said in #27 , the issue is more about what it would mean to have a handle to a stream in a JS environment (meaning that another library would be able to use the handle and underlying stream properly), rather than just providing a wrapper for the upstream object. 
A possible solution would be consider these just as Node streams (which are not really MediaStreams neither), but then we still need to wait for https://github.com/napi-rs/napi-rs/discussions/1483 .
In any case, there is no simple solution right now unfortunately, what would be your use case?

justin0mcateer:
I don't actually want to handle the MediaStream directly in the JavaScript, I need to hand it off to another library (WebRTC) that is expecting specifically a MediaStream (not a NodeJS Stream).

orottier:
Hey @justin0mcateer, thanks for bringing this up.

The problem is, like @b-ma states, there is no such thing as the `MediaStream` in NodeJs. So even if we would provide an implementation here, it will not be compatible with other libraries. (The w3c only specifies the API surface and relation to other APIs such as webaudio and webrtc, not the inner workings). See e.g. https://github.com/node-webrtc/node-webrtc/issues/728#issue-1214108642 for the issues that will arise.

On the web-audio-api-rs side of things I acknowledge this issue, and we solve it by provinding a real-time safe MediaStream shim that allows you to extract the audio samples ( https://docs.rs/web-audio-api/latest/web_audio_api/media_streams/struct.MediaStreamTrack.html#method.iter ) and it allows you to construct an instance from raw audio data ( https://docs.rs/web-audio-api/latest/web_audio_api/media_streams/struct.MediaStreamTrack.html#method.from_iter ). 

The NodeJs ecosystem should provide the same so application authors will be able to glue libraries such as node-web-audio-api and node-webrtc manually. So perhaps this issue could be rewritten to "How to extract raw audio without resorting to AudioWorkletProcessor"?

[...]

PR #105, opened 2024-04-07
Fix type errors

b-ma:
Almost got rid of the type error issues

#### before
```
RESULTS:
  - # pass: 6748
  - # fail: 791
  - # type error issues: 24
```

#### after
```
RESULTS:
  - # pass: 6813
  - # fail: 743
  - # type error issues: 5
```

Remaining one are linked to:
- the-mediastreamaudiodestinationnode-interface/ctor-mediastreamaudiodestination.html (not relevant for now, cf. #91)
- the-audionode-interface/audionode-disconnect.html (cf. https://github.com/orottier/web-audio-api-rs/issues/471)

### DIFF
--- a/.scripts/wpt-harness.mjs
+++ b/.scripts/wpt-harness.mjs
@@ -124,7 +124,7 @@ const reporter = {
   },
   pass: message => {
     numPass += 1;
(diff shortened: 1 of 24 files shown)


orottier (approved):
Very nice. LGTM

PR #108, opened 2024-04-14
Fix - decodeAudioData in another thread

b-ma:
@orottier do you think this approach is ok or do I miss something? (seems to work well but...)

todo
- [ ] sanitize output error, cf. "Let error be a DOMException whose name is [EncodingError](https://webidl.spec.whatwg.org/#encodingerror)."


orottier:
Hey this is perfectly fine because I intend to write almost identical code for https://github.com/orottier/web-audio-api-rs/issues/423. After we land that one we can clean up your code a bit.

b-ma:
Ok perfect, thanks!

Issue #111, opened 2024-04-14
Bind the ScriptProcessorNode


orottier:
It exists now: https://docs.rs/web-audio-api/1.0.0-rc.4/web_audio_api/node/struct.ScriptProcessorNode.html

b-ma:
Yup, this is in the short list :) but I need to review/generalize how thread safe functions are stored / handled first

b-ma:
Done in #121

PR #113, opened 2024-04-15
Refactor - generalize thread safe function store over different event types

b-ma:
Required for #111

The gut of the PR is here: https://github.com/ircam-ismm/node-web-audio-api/pull/113/files#diff-14aa34af7493adadd4e38062142862cf9df89cccec3fa02f78dc81ba0787cfd5

Any hint / idea on how to simplify / clean that is welcome!

### DIFF
--- a/generator/rs/audio_nodes.tmpl.rs
+++ b/generator/rs/audio_nodes.tmpl.rs
@@ -1,6 +1,7 @@
 use napi::*;
 use napi_derive::js_function;
 use web_audio_api::node::*;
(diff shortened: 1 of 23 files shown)


b-ma:
hum does not work as expected...

b-ma:
Actually, I'm thinking we could probably get rid of this store thing altogether...

b-ma:
Killing in favor of #116 that was a silly pile of complicated stuff...

Issue #80, opened 2024-02-06
Review `AudioBuffer::getChannelData` implementation


b-ma:
I think this inhibits some undefined behaviour - theoretically.

You are asking the compiler for a mutable reference of the AudioBuffer, but then this information is 'erased' via the unsafe slice cast. This means if you call the method twice and mutate one of the copies, the compiler could assume the other copy will not be affected.

I don't have a clear solution though (perhaps return the same TypedArray via `set_property`??) - should we add an issue to address this later?

_Originally posted by @orottier in https://github.com/ircam-ismm/node-web-audio-api/issues/79#issuecomment-1930401582_

b-ma:
This is completely unusable, the underlying memory can be reused at any moment, so keeping around the returned Float32 will end up being full of garbage...

A possible solution would be to return an immutable reference, but this will probably break some wpt tests

b-ma:
## Acquire the content

Spec here: https://wiki.mozilla.org/User:Roc/AudioBufferProposal#Implementation_Sketch

Note that "neutering" is called "detach" now:
- https://github.com/nodejs/node-addon-api/issues/541
- https://tc39.es/ecma262/#sec-detacharraybuffer

PR #114, opened 2024-04-15
Feat - some ScriptProcessorNode tail

b-ma:
First (...failed) attempt

### To make it short
- I think the rust API is nice/idiomatic/relevant as it is
- I have no idea how I can deal with it

### Most simplified example

relevant code is here: <https://github.com/ircam-ismm/node-web-audio-api/compare/feat/script-processor?expand=1#diff-f2a3e90afd7d940e52b634347f9a1e12ee21402dafa586f92c960ad13fa052fcR146>

In short:

```rs

node.set_onaudioprocess(move |e| {
    tsfn.call(Ok(e), ThreadsafeFunctionCallMode::Blocking);
});
```

### Problem

```
error[E0521]: borrowed data escapes outside of closure
   --> src/script_processor_node.rs:171:17
    |
169 |             node.set_onaudioprocess(move |e| {
    |                                           -
    |                                           |
    |                                           `e` is a reference that is only valid in the closure body
    |                                           has type `&'1 mut AudioProcessingEvent`
170 |                 // let event = WebAudioEventType::from(e);
171 |                 tsfn.call(Ok(e), ThreadsafeFunctionCallMode::Blocking);
    |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |                 |
    |                 `e` escapes the closure body here
    |                 argument requires that `'1` must outlive `'static`
```

\o/

### DIFF
--- a/examples/script-processor.mjs
+++ b/examples/script-processor.mjs
@@ -5,7 +5,6 @@ const audioContext = new AudioContext({ latencyHint });
 
 const sine = new OscillatorNode(audioContext);
 sine.frequency.value = 200;
-sine.connect(audioContext.destination);
 sine.start();
 
 const scriptProcessor = new ScriptProcessorNode(audioContext);
@@ -14,5 +13,6 @@ scriptProcessor.addEventListener('audioprocess', e => {
   console.log(e.inputBuffer);
   console.log(e.outputBuffer);
 });
-scriptProcessor.connect(audioContext.destination);
+
+sine.connect(scriptProcessor).connect(audioContext.destination);
 
--- a/examples/script-processor.mjs
+++ b/examples/script-processor.mjs
@@ -1,0 +1,18 @@
+import { AudioContext, OscillatorNode, ScriptProcessorNode } from '../index.mjs';
+
+const latencyHint = process.env.WEB_AUDIO_LATENCY === 'playback' ? 'playback' : 'interactive';
+const audioContext = new AudioContext({ latencyHint });
+
+const sine = new OscillatorNode(audioContext);
+sine.frequency.value = 200;
+sine.connect(audioContext.destination);
+sine.start();
+
+const scriptProcessor = new ScriptProcessorNode(audioContext);
+scriptProcessor.addEventListener('audioprocess', e => {
+  console.log(e.playbackTime);
+  console.log(e.inputBuffer);
+  console.log(e.outputBuffer);
+});
+scriptProcessor.connect(audioContext.destination);
+
--- a/examples/script-processor.mjs
+++ b/examples/script-processor.mjs
@@ -1,5 +1,5 @@
-import { AudioContext, GainNode, OscillatorNode, ScriptProcessorNode } from '../index.mjs';
-import { kNapiObj } from '../js/lib/symbols.js'
+import { AudioContext, OscillatorNode, ScriptProcessorNode } from '../index.mjs';
+import { sleep } from '@ircam/sc-utils';
 
 const latencyHint = process.env.WEB_AUDIO_LATENCY === 'playback' ? 'playback' : 'interactive';
 const audioContext = new AudioContext({ latencyHint });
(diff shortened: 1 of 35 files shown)


orottier:
Right, the signature `F: FnMut(&mut AudioProcessingEvent)` is simply not compatible with the ThreadSafeFunction way of working.

Even if we would change it to `F: FnMut(Arc<Mutex<AudioProcessingEvent>>)` (so you can clone it out while still being able to mutate) it would not work because right after the event handler is run the buffer is shipped back to the render thread. So you will not be in time with the actual mutation.

Only solution I can come up with now is to provide a completely different (hidden) API. In which the AudioProcessingEvent contains a oneshot channel to the ScriptRenderer. I will have to sketch that out a bit before knowing it will really work.

b-ma:
> Right, the signature F: FnMut(&mut AudioProcessingEvent) is simply not compatible with the ThreadSafeFunction way of working.

Ok thanks, I can stop trying to imagine fancy workarounds :)

If `ThreadsafeFunctionCallMode::Blocking` does what it pretends, maybe the first solution `F: FnMut(Arc<Mutex<AudioProcessingEvent>>)` could work no? I will make a dirty hack to test it, I will let you know

orottier:
I think we established earlier that Blocking does not execute immediately, but its worth the try. 
Im working on an implementation now that suits both rust and node, hooking into the Drop of the processing event. Stay tuned

b-ma:
> I think we established earlier that Blocking does not execute immediately, but its worth the try.

Right, seems I misunderstood the blocking thing, this seems to refer only to how the callback is added to the event queue.
There is still more than one reference when a try to unwrap the Arc<Mutex> to its inner value...

> Im working on an implementation now that suits both rust and node, hooking into the Drop of the processing event. Stay tuned

Ok cool, let me know!

orottier:
Really cool you managed to get it to work. Quite the milestone to finally have custom JS audio processing now.
I need a bit more time to look at all of it. I'm a bit worried about your new ThreadSafeFunction implementation, what are the exact changes and how do we keep it up to date over time?
I will merge the new Rust API for the processing event. I think it's fine in the new way.

examples/script-processor.mjs
```diff
@@ -0,0 +1,27 @@
...
+
+const latencyHint = process.env.WEB_AUDIO_LATENCY === 'playback' ? 'playback' : 'interactive';
+const audioContext = new AudioContext({ latencyHint });
+
+const sine = new OscillatorNode(audioContext);
+sine.frequency.value = 200;
+
+const scriptProcessor = new ScriptProcessorNode(audioContext);
+const buffer = new Float32Array(scriptProcessor.bufferSize);
+scriptProcessor.addEventListener('audioprocess', e => {
```
orottier, line 12: We should be able to modify the samples without an intermediate buffer no?
First just copy the input to output, then something like

    const channelData = audioBuffer.getChannelData(channel);

    for (let sample = 0; sample < channelData.length; sample += 1) {
        channelData[sample] +=  Math.random() * 2 - 1;
    }

b-ma:
> I'm a bit worried about your new ThreadSafeFunction implementation, what are the exact changes and how do we keep it up to date over time?

Yup that may be an issue, but several mitigating points:
- I didn't wrote it myself :)
- the guy from napi-rs himself point to this solution here: https://github.com/napi-rs/napi-rs/discussions/1334
- this is still in use in the [lightningcss](https://github.com/parcel-bundler/lightningcss) project which is quite popular and maintained, cf. https://github.com/parcel-bundler/lightningcss/blob/master/napi/src/threadsafe_function.rs (this even seems to be part of this crate https://docs.rs/lightningcss-napi/latest/lightningcss_napi/, but not exposed)
- the code appear to be rather low level napi code so it should be rather stable

examples/script-processor.mjs
b-ma: Yes indeed, this works as well. I just didn't want to pile up possible issues for this first test as our `getChannelData` is not the most reliable thing for now, cf. #80

b-ma:
Just to say, this will all be probably dropped and rewritten because it is based on dead branch

No idea yet on how  to get rid of the alternative implementation though... However I think it's good anyway to put a nail into the guts because I'm pretty sure `AudioWorklet` (to work properly) will require importing and binding stuff like `node.h`, `lib_uv.h` or `v8.h`

orottier:
Alright, let's go with this version for now then. I will release the new owned API today so this branch can be merged

orottier:
`v1.0.0-rc.6` has the API this PR requires.

PR #116, opened 2024-04-19
Refactor - review events

b-ma:
Cleaner alternative compared to #113, can basically end up with removing the ThreadsafeFunction store

Pretty happy with that so far (no more `std::thread::sleep(std::time::Duration::from_millis(1));`), let's see where it goes...

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,6 @@ crate-type = ["cdylib"]
 [dependencies]
 napi = {version="2.15", features=["napi9", "tokio_rt"]}
 napi-derive = "2.15"
-uuid = {version="1.6.1", features = ["v4", "fast-rng"]}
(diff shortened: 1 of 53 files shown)


orottier:
From a distance it looks pretty clean. I'm also happy the store, the uuids and the sleeps are not needed.
This only contains one event still? Will you add the others or because they do not have member fields it is not necessary?

b-ma:
>  I'm also happy the store, the uuids and the sleeps are not needed.

Yup me too, really more clean this way

> Will you add the others or because they do not have member fields it is not necessary?

Actually the only events defined in the spec are:
- `AudioRenderCapacityEvent`
- `OfflineAudioCompletionEvent`
- `AudioProcessingEvent`
Other ones are simple raw events with just a `type` member

orottier:
Okay, great!

Issue #28, opened 2023-08-01
AudioWorklet support


jon-myers:
calling `.audioWorklet` on an initialized `AudioContext` returns `undefined`. Is working with web audio audio worklets supported?

b-ma:
Hey, no the AudioWorklet API is not yet supported unfortunately as few other nodes. Hopefully it will be in some future but I can't give any roadmap nor due date. 

 I will had some notes on that in the README

jampekka:
I'd be interested in looking into this. The biggest hurdle is probably to get the AudioWorkerProcessor to run in the web-audio-api-rs audio rendering thread. I think this could be possible by spawning a new napi::Env in the rendering thread, call Env::run_script on that and implement AudioWorkletProcessor in the rust side that then calls the JS AudioWorkletProcessor defined in Env::run_script.

I'm very new to rust and node.js internals, so this may not be the way to go. Could somebody more familiar with these comment if this architecture is worth pursuing?

orottier:
Hi @jampekka that would be very welcome. Your suggestion makes sense and I have looked into this before, but I was unable to get an env set up on the render thread. It is obviously not allowed to use the Env of the main thread, but nowhere in the n-api docs I could find a way to construct a new one.

I will investigate further and report back in this issue. If you do the same, that would be great

b-ma:
> but nowhere in the n-api docs I could find a way to construct a new one.

I have made a bit of research and found this https://docs.rs/napi/latest/napi/attr.module_init.html which seems to correspond to `NAPI_MODULE_INIT` in https://nodejs.org/api/n-api.html, which seems to be the macro that creates a new env 

> The parameters env and exports are provided to the body of the NAPI_MODULE_INIT macro.

Maybe this can help?

orottier:
Hey @b-ma, I read those docs but I think that is not related to the actual executor. But more for the module loading (require ...) and running a one-time initialization method on the module. But I may be mistaken.

I'm simultaneously looking into https://nodejs.org/api/vm.html which may be interesting provided it does not share the event loop with the main thread
> A common use case is to run the code in a different V8 Context. This means invoked code has a different global object than the invoking code.

Another option is to spawn a worker https://nodejs.org/api/worker_threads.html
> The node:worker_threads module enables the use of threads that execute JavaScript in parallel. ...
> Workers (threads) are useful for performing CPU-intensive JavaScript operations.

[...]

PR #124, opened 2024-05-03
Work in progress for AudioWorkletNode via Worker thread

orottier:
Many many todo, but it works now for single-channel source nodes! `examples/audio-worklet.mjs`
Relates to #28 

Todo
- [ ] pass inputs
- [ ] pass all outputs
- [ ] pass parameterData
- [ ] return the tailTime result to rust
- [ ] pass the currentTime, currentFrame, sampleRate
- [ ] use the AudioWorkletNodeOptions
- [ ] stdout of the Worker is lost somehow (console.log does not work)
- [ ] do not spawn a Worker thread per Node
- [ ] conform to the spec (needs `context.addModule`, only pass the name in the `AudioWorkletNode` ctor)
- [ ] make the messagePort work. This is tricky because we need to spin the event loop instead of blocking it permanently
- [ ] set thread priority in Worker (can be done with thread_priority crate)
- [ ] ...


index.d.ts
```diff
@@ -1,25 +1,6 @@
...
-    export import GainNode = globalThis.GainNode;
-    export import IIRFilterNode = globalThis.IIRFilterNode;
-    export import OscillatorNode = globalThis.OscillatorNode;
-    export import PannerNode = globalThis.PannerNode;
-    export import StereoPannerNode = globalThis.StereoPannerNode;
-    export import WaveShaperNode = globalThis.WaveShaperNode;
-}
+/* auto-generated by NAPI-RS */
+
+export function runAudioWorklet(): void
```
orottier, line 6: not sure what happened to this file and the other index.js

b-ma:
Cool! I will have a look soon

generator/rs/lib.tmpl.rs
```diff
@@ -1,7 +1,63 @@
...
+                })
+            .unwrap()
+        };
+
+        let mut output_channels = env.create_array(0)?;
+        output_channels.insert(output_samples)?;
+        let mut outputs = env.create_array(0)?;
+        outputs.insert(output_channels)?;
+
+        let _ret: bool = process.call3(
```
orottier, line 53: I think we need apply3 to actually bind `this`

index.d.ts
b-ma: I think it is due to the fact the you used the `#[napi]` macro which is really high level.

I will have a look to align this with the rest of the codebase logic, should not be complicated

generator/rs/lib.tmpl.rs
b-ma: I guess we will need monkey patching on the JS side to have the right context / this when `run_audio_worklet` is called

b-ma:
Just made small change to remove the hardcoded path and have it work on my machine, this is really nice! Great work!

generator/rs/lib.tmpl.rs
orottier: My idea is that we keep a list of all worklet processors on the rust side, and use some extra fields in the message channel to select the proper processor for the render call.

Cargo.toml
```diff
@@ -10,10 +10,11 @@ version = "0.20.0"
 crate-type = ["cdylib"]
 
 [dependencies]
+crossbeam-channel = "0.5.12"
 napi = {version="2.15", features=["napi9", "tokio_rt"]}
 napi-derive = "2.15"
-web-audio-api = "=0.44.0"
-# web-audio-api = { path = "../web-audio-api-rs" }
+# web-audio-api = "=0.44.0"
+web-audio-api = { path = "../web-audio-api-rs" }
```
orottier, line 17: You will need the branch `feature/param-keys` for now

src/audio_worklet_node.rs
```diff
@@ -0,0 +1,260 @@
...
+                .env
+                .unwrap::<NapiOfflineAudioContext>(&js_audio_context)?;
+            let audio_context = napi_audio_context.unwrap();
+            AudioWorkletNode::new::<NapiAudioWorkletProcessor>(audio_context, options)
+        }
+        &_ => panic!("not supported"),
+    };
+
+    /* TODO expose parameters in JS
+    dbg!(native_node.parameters());
```
orottier, line 176: @b-ma I could maybe use your help to expose the AudioParams to javascript here.
If you uncomment the rust lines and run the example you will see that the bitcrusher now works with automations, but obviously we should be able to control the params from the JS side.

generator/rs/lib.tmpl.rs
b-ma: Makes sens, and your solution is the most straightforward indeed, from the documentation I was afraid we could only pass the original `Context` to `apply`.

src/audio_worklet_node.rs
b-ma: Ok sure, I will have a look

Just want to fix this `index.js|ts` files issue first

index.d.ts
b-ma: Ok really strange, the changes are seems to be good but there is some sort of cache that makes these 2 files reappear (or erased) by napi constantly. But if I clone the repo again, everything is good...

b-ma: Just spend 1 hour and half fiddling on that but cannot figure out where this comes from, so I just tried to go brute force and restart from a fresh clone... but the cache seems to be linked on the directory name, that's GRRRR!

If I just clone with another directory name, everything is fine...

b-ma: Ok quite insane, step I made to have it work properly (seems the reboot was the big deal, because other steps alone didn't work but not really sure neither):

```
rm -Rf target/
rm -Rf node_modules/
cargo cache -a # using https://crates.io/crates/cargo-cache
rm -Rf ~/.npm/_cacache/

# reboot computer now!

npm install 
npm run build
```

And everything behaved as expected finally...

b-ma:
[WARNING] make sure to check https://github.com/ircam-ismm/node-web-audio-api/pull/124#discussion_r1589944179 to not reintroduce the `index.js` file and changes in `index.d.ts`

b-ma:
I can advance on 
> - conform to the spec (needs context.addModule, only pass the name in the AudioWorkletNode ctor)

next, can be done on JS side rather simply I think

orottier:
Thanks! Interesting stuff happening regarding the index.js. I will take care

> can be done on JS side rather simply I think

Okay that would be helpful!

js/AudioWorkletNode.js
```diff
@@ -0,0 +1,121 @@
...
+      if (arguments.length < 2) {
+        throw new TypeError(`Failed to construct 'AudioWorkletNode': 2 arguments required, but only ${arguments.length} present`);
+      }
+
+      if (!(context instanceof jsExport.BaseAudioContext)) {
+        throw new TypeError(`Failed to construct 'AudioWorkletNode': argument 1 is not of type BaseAudioContext`);
+      }
+
+      // parsed version of the option to be passed to NAPI
+      const parsedOptions = {};
```
orottier, line 37: @b-ma If you have the time I would really appreciate if you could sanitize the options and use them on the Rust side. We can then remove the default option set from https://github.com/ircam-ismm/node-web-audio-api/pull/124/files#diff-5221df6d2a274b6c20b0dd35dbfbfd3a4a78ff08a5cdfbc032edd0e7ae72ef15R182
The type `AudioWorkletNodeOptions<()>` is still suitable because we move the processorOptions to the Worker on the JS side already.

b-ma: Ok, the thing is done on the JS side, to be continued

(I also have a rather interesting prototype for the `addModule` thing in the pipes :)

b-ma: Done!

js/AudioWorkletGlobalScope.js
```diff
@@ -0,0 +1,91 @@
...
+
+  console.log(event.cmd);
+
+  switch (event.cmd) {
+    case 'node-web-audio-api:worklet:add-module': {
+      const { code } = event;
+      const func = new Function('AudioWorkletProcessor', 'registerProcessor', code);
+      func(AudioWorkletProcessor, registerProcessor);
+
+      // @todo - send back parameterDescriptors
```
b-ma, line 69: Need to implement some request / response mechanism, then using it to properly close the AudioWorkletGlobalScope will be straightforward

b-ma: > then using it to properly close the AudioWorkletGlobalScope will be straightforward

Or maybe not, to be investigated...

examples/worklets/bitcrusher.js
```diff
@@ -0,0 +1,77 @@
...
+        }, {
+            name: 'frequencyReduction',
+            defaultValue: 0.5,
+            minValue: 0,
+            maxValue: 1
+        }];
+    }
+
+    constructor(options) {
+      process.stdout.write(`++ in constructor: ${JSON.stringify(options, null, 2)}\n`);
```
b-ma, line 17: Don't really get why these prints are swallowed, is this a regression ?

js/AudioWorkletGlobalScope.js
```diff
@@ -0,0 +1,91 @@
...
+      const { name, processorOptions, messagePort } = event;
+      const ctor = nameProcessorCtorMap.get(name);
+
+      register_params(ctor.parameterDescriptors ?? []);
+
+      processorOptions[kMessagePort] = messagePort;
+      const instance = new ctor(processorOptions);
+
+      // @todo - enable multiple processors
+      globalThis.proc123 = instance;
```
b-ma, line 81: Should we use some kind of uuid to match the rust processor and the related JS instance?

examples/worklets/bitcrusher.js
orottier: Yeah this used to work (via console.log even) when we started using `setImmediate` in the worker loop

js/AudioWorkletGlobalScope.js
orottier: Yeah I can set up an incrementing ID on the Rust side, and then we use some sort of Map to store the JS processors here?

orottier:
Okay this setup looks pretty good!
One issue: if I run npm:build it overrides a change in `js/BaseAudioContext.js` so probably you need to update the generator?
I will look into the ID generation now

orottier:
Sorry, I did not come really far and I have to be away from keyboard for the rest of today.
My work is over at commit ﻿25743a6c1d8862251e5c6d88119a749e112a1add (not on this branch, you can just fetch it) but it hangs because of the message channel stuff I think, I reordered some lines in js/AudioWorkletNode.js

b-ma:
> One issue: if I run npm:build it overrides a change in js/BaseAudioContext.js so probably you need to update the generator?

Indeed, this is fixed

> I have to be away from keyboard for the rest of today.

Same here, have a nice day!

js/AudioWorkletGlobalScope.js
b-ma: Or the other way around, don't know which one is the more straightforward

examples/worklets/bitcrusher.js
b-ma: The logs just re-appeared on my side, even with `console.log`, weird...

b-ma:
> but it hangs because of the message channel stuff I think, I reordered some lines in js/AudioWorkletNode.js

This is because in `AudioWorkletGlobalScope.js`, `register_params(parameterDescriptors);` is called within `context.audioWorklet[kCreateProcessor]`, which is too late considering the lines you reordered `AudioWorkletNode.js` (and probably wrong actually). I think the `NapiAudioWorkletProcessor::parameter_descriptors` just waits for data that has not been sent yet.

I just moved `register_params(parameterDescriptors)` inside `registerProcessor` in [f2ad75c](https://github.com/ircam-ismm/node-web-audio-api/pull/124/commits/f2ad75cfaf37159f5b096b4b21cfb2ae57030ccf) (which is very probably where it should live), this should solve the problem I think.

------ 

I'm just realizing we should probably ensure the JS process exists in `run_audio_worklet` because it seems we can run into some sort of race condition between the rust audio thread and the worklet thread, i.e. maybe something like:

```rs
let proc = global.get_named_property::<JsUnknown>("proc123")?;
if proc.get_type() == ValueType::Undefined {
   // js processor not ready yet
   return ctx.env.get_undefined()
}

let proc = proc.coerce_to_object()?
// ...
```

Another small thing I have seen in your changes: it would be probably more simple as first step to just store the processor in `globalThis` rather than in a `WeakMap`, e.g. :

```js
globalThis[`processor-${id}`] = instance;
```

I'm not sure if and how we can access Maps from Napi, and `WeakMap` create weak references to their keys not to their values, so this will have the same effect anyway. 
This will probably be an problem for garbage collection but maybe we can keep this problem for later

orottier:
Thanks for all the info.

I think we can actually get rid of `register_params`. I needed it because only the Worker knew about the audioParamDescriptors, but I think we have this info on the main thread already because of your work (`kWorkletParamDescriptorsMap`). This means we don't need a back and forth between the worker anymore. We can request the map from the AudioWorkletNode js constructor, pass it to rust, and call `send_recv_pair2().0.send` from the `NapiAudioWorkletNode` constructor

I hope I can find some time in the next days to address this, but feel free to take a stab yourself.

> I'm just realizing we should probably ensure the JS processor exists in run_audio_worklet 

Yeah, good idea probably

> just store the processor in globalThis

Sure!

b-ma:
> I think we can actually get rid of register_params.

Sure, I will do this

orottier:
Looking good!

src/audio_worklet_node.rs
```diff
@@ -0,0 +1,379 @@
...
+    current_time: f64,
+    // AudioWorkletGlobalScope currentFrame
+    current_frame: u64,
+    // AudioWorkletGlobalScope sampleRate
+    sample_rate: f32,
+    // tail_time return value
+    tail_time: Arc<AtomicBool>,
+}
+
+unsafe impl Send for ProcessorArguments {}
```
orottier, line 33: This is no longer necessary

src/audio_worklet_node.rs
```diff
@@ -0,0 +1,379 @@
...
+        let (send, recv) = crossbeam_channel::unbounded();
+        Mutex::new((Some(send), Some(recv)))
+    })
+}
+
+pub(crate) struct SendItem2(Vec<AudioParamDescriptor>);
+
+// channel from worker to main
+#[allow(clippy::type_complexity)] // will refactor later
+pub(crate) fn send_recv_pair2() -> &'static (Sender<SendItem2>, Receiver<SendItem2>) {
```
orottier, line 57: The Sender must be wrapped in a Mutex to avoid mix-ups in concurrent access (not entirely sure if that is possible from nodejs, but let's keep it safe)
The MutexGuard must be held from the `send` call to after the  `AudioWorkletNode::new` call

src/audio_worklet_node.rs
```diff
@@ -0,0 +1,379 @@
...
+        let outputs: &'static [&'static [&'static [f32]]] = unsafe { std::mem::transmute(outputs) };
+        let params: Vec<(&'static str, &'static [f32])> = params
+            .keys()
+            .map(|k| {
+                let label = unsafe { std::mem::transmute(k) };
+                let value = unsafe { std::mem::transmute(&params.get(k)[..]) };
+                (label, value)
+            })
+            .collect();
+        let item = ProcessorArguments {
```
orottier, line 367: Include the incrementing ID here

src/audio_worklet_node.rs
```diff
@@ -0,0 +1,379 @@
...
+    let mut global = ctx.env.get_global()?;
+
+    // fill AudioWorkletGlobalScope
+    global.set_named_property("currentTime", current_time)?;
+    global.set_named_property("currentFrame", current_frame)?;
+    global.set_named_property("sampleRate", sample_rate)?;
+
+    // Make sure the processor exists, might run into race conditions between
+    // Rust Audio thread and JS Worker thread
+    let processor = global.get_named_property::<JsUnknown>("proc123")?;
```
orottier, line 91: Utilize the incrementing ID here to select to correct processor

orottier:
I'm on mobile so I left some note-to-self. Not sure when I will be able to pick them up so feel free to pick up things too!

I guess the ability to address multiple processors in the same Worklet is the only thing really left beside pesky details such as real-time safety, destructors and some code cleanup :D

b-ma:
Ok I'll try to advance on these on my side, I'm in a flow :)

b-ma:
This still doesn't work with 2 worklets, I'm having a look

orottier:
I added another to-do regarding the communication channels when multiple audio contexts run at the same time (could be a single online and a few offline ones for example). I have a bit of work in progress for that but probably will need some of your help later! 

Regarding this:

> maybe use the `before_drop` method to notify the Js Worker that the JS processor can be dropped

That could be a good idea indeed. The before_drop hook is not publicly exposed in the audioworklet processor, but I could change that. Or we use a similar mechanism just hooking onto the actual `Drop`

b-ma:
> This still doesn't work with 2 worklets, I'm having a look

There is an issue here I'm not sure how to handle, when the second worklet is created, there is a panic due to:

```rs
// audio_worklet_node.rs, line 270
let send = send_recv_pair().lock().unwrap().0.take().unwrap();
```

Which actually make sens, should we clone it instead of take?

orottier:
Yes, definitely

orottier:
Please have a look at my latest commit. I somehow needed this because otherwise the Worker would panic 50% of the times when booting. With this change it seems to work fine having multiple AudioWorkletNodes ¯\\\_(ツ)\_/¯

orottier:
> garbage collect the Js processor in the Worker when finished processing

I guess this is done with the lastest commit, but now I added a task

> close the AudioWorkletGlobalScope properly so a script can terminate

Because there are still some things preventing a script to terminate it seems.

b-ma:
\o/ :)

b-ma:
> Because there are still some things preventing a script to terminate it seems.

I'll check that today, we just need to properly close the `Worker`, when `AudioContext::close` is called, or `OfflineAudioContext::startRendering` has ended I think

b-ma:
Other todos:

- [ ] Error to investigate when 2 worklets in same script:
```
node-web-audio-api:worklet:create-processor


node:internal/event_target:1096
  process.nextTick(() => { throw err; });
                           ^
TypeError [Error]: Cannot convert undefined or null to object
```

- [ ] do not expose `run_audio_worklet`

orottier:
Regarding the task `the comms channel approach does not work if there are multiple AudioContexts running at the same time, we need to bind the channels to the actual BaseAudioContext instance`, I have the following plan:

The `AudioParamDescriptorsChannel` is sound. It is not an issue that this channel is per application instead of per AudioContext, as concurrent calls cannot happen because of the Mutex around the sender. After the send/recv handover of the descriptors is done, the channel can be reused by another context for another node.

The application-wide `ProcessCallChannel` is not sound when multiple contexts run concurrent because it might be that a process call is delegated to the wrong Worker instance. This means we need to store it per BaseAudioContext:
- `run_audio_worklet` needs to receive a JS argument containing the Receiver and should substitute `process_call_channel().recv` with the unwrapped Receiver
- hence we need a JsObject wrapping the Receiver living in the Worker globalThis
- the Sender is part of the NapiAudioWorkletProcessor, and must be equal (cloned) for all processor in the same context
- hence the Sender must be a field of NapiAudioContext and NapiOfflineAudioContext, so the NapiAudioWorkletProcessor can retrieve it during the constructor call

Not sure how this can be implemented best. Maybe in the Napi(Offline)AudioContext constructor, and then passed down into the Worker eventually. Do you have suggestions?

b-ma:
Hum right, your plan sounds good to me, I really don't see how to achieve that without making the sender doing the whole Rust -> JS main -> JS worklet path

Let me know if you need some help for the JS side

b-ma:
> Unsafety checks: what happens if we write to the inputs? what happens if we store the outputs in globalThis and use them after the processor has been dropped? How can we utilize env.freeze/FrozenArray?

Tried a naive `let _ = samples.freeze()?;` but got `Error "Cannot freeze array buffer views with elements"` which is aligned with [mdn](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze): "A [TypedArray](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypedArray) or a [DataView](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/DataView) with elements will cause a [TypeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError), as they are views over memory and will definitely cause other possible issues"

(see also https://es.discourse.group/t/make-readonly-typedarray/712)

Or we might use https://docs.rs/napi/latest/napi/struct.Env.html#method.create_arraybuffer_with_data just as in `wave_shaper_node::get_curve`, but as there is a `slice.to_vec()` involved... I guess it allocates

Or maybe create a clone of the underlying `ChannelData` especially for Worklet, so that if it is accidentally written it wont pollute the "real" values? (but with all the black magic involved I really don't know if it would work)

Or live with the fact that it is possible, and that people should know what they are doing when writing such code

(and I'm out of ideas :)

orottier:
Hm okay, I also tried to freeze the underlying `array_buffer` but to no avail.
In any case I'm fine with a disclaimer "Your AudioWorkletProcessor should be well behaved and you should not run untrusted scripts" and ignore the issue for now. I need to read more about it anyway to really understand what we are trying to mitigate here.

I will try to see how far I get with the `Sender` propagation!

b-ma:
> There is a big TODO, how do we get a JsObject wrapping the NapiAudioContext.1 (ProcessCallChannel) value into the Worker? Can such a thing even be sent via message passing? It obviously cannot be serialized.

Damn yes, I was hoping we could transfert it just like we do for the `port` but looking more closely, it's very probably not an option yet https://github.com/nodejs/node/issues/37080

Could we keep something similar to the current design and filter messages in `run_audio_worklet` using some kind of `context_id`? Maybe with storing a napi-wrapped clone of the receiver in each Worker `globalThis`?

js/AudioWorkletGlobalScope.js
```diff
@@ -11,7 +11,7 @@ const kWorkletInputs = Symbol.for('node-web-audio-api:worklet-inputs');
 const kWorkletOutputs = Symbol.for('node-web-audio-api:worklet-outputs');
 const nameProcessorCtorMap = new Map();
 const paramDescriptorRegisteredMap = new Map();
-// const processorIdMap = new WeakMap(); // instance, uuid
+const workletId = 0; // TODO how to receive from the Rust side?
```
orottier, line 14: I could use your help here @b-ma. The workletId is in `Napi(Offline)AudioContext.worklet_id()` and I need it here :)

b-ma: Ok I'll have a look

b-ma: Done!

js/AudioWorkletGlobalScope.js
```diff
@@ -0,0 +1,138 @@
...
+  if (obj === null || obj === undefined) {
+    return false;
+  }
+  return typeof obj[Symbol.iterator] === 'function';
+}
+
+function runLoop() {
+  // console.log(workerData.workletId);
+  // block until we need to render a quantum
+  run_audio_worklet(workerData.workletId);
```
orottier, line 32: Probably use the const, unpacked value here?
```suggestion
  run_audio_worklet(workletId);
```

b-ma: (sorry, I don't have the reflex of github suggestions yet, anyway this is done!)

orottier:
Alright I'm running out of ideas for further to-dos. How about you? How are we doing on the WPT side?

b-ma:
In any case, I think the big picture is there

> How about you? How are we doing on the WPT side?

Not quite there yet :) there a lot of small things, but I think they are mostly things that can handled on JS side. I need to dig into why the tests fail before we got a picture on this point.

I will continue digging into the wpt a little bit, let you know if there something I don't know how to deal with

...And I got a Seg Fault too when running the wpt... Which is a little bit annoying, but maybe that was due to another issue, e.g. executing trash instead of actual code, I don't know, let's see

Thanks in any case, that's really cool to have this (and not so much a hack in the end :) !

orottier:
Alright, well then I would like to congratulate ourselves with this great collaboration!

I will play around a bit more, maybe refactor/clean a bit on the rs side

orottier:
Humm, I may have some bad news. I was writing the SAFETY comments and realized a big omission in the Rust <> Worker handover. We never await the result. So the rust processor sends a command to the Worker, and then immediately continues. However it should wait until the Worker has finished the process call of that node. This might explain the occasional segfault, and maybe some WPT test results.

I have this fixed now but the load seems quite high, there's a framedrop every now and then..

orottier:
The call into the Worklet processor seems to take 0.05 - 0.15 milliseconds (bitcrush call) which seems not too bad (a few percent of the budget). So I assume something in the event loop is sleeping/idling. The funny thing is, if I log many things, I do not get framedrops. As if IO is keeping the event loop more active.. Enough for tonight, I guess

b-ma:
Nice catch, but there are weird things happening indeed: 
- when I run the `examples/audio-worklet.mjs` now it only starts properly 20-25% of the time and stays stuck at beginning the rest the time
- the script does not exit properly anymore

Another (maybe related) thing I was thinking, do we, and if "no", can we ensure that the `run_audio_worklet` function is called only once per render quantum even if we have multiple worklets? So that `setImmediate` is called only when the graph has finished rendering?

Let me know if I can help testing things on my side

b-ma:
> - when I run the `examples/audio-worklet.mjs` now it only starts properly 20-25% of the time and stays stuck at beginning the rest the time

Interestingly this does not happen with `examples/audio-worklet-shared-array-buffer.mjs` which has only 1 worklet running

b-ma:
> Another (maybe related) thing I was thinking, do we, and if "no", can we ensure that the run_audio_worklet function is called only once per render quantum even if we have multiple worklets? So that setImmediate is called only when the graph has finished rendering?

What about something like:
```rs
// Wait for an incoming command
// better to starve JS worklet event loop
while let Ok(msg) = process_call_receiver(worklet_id).recv_timeout(Duration::from_micros(500)) {
    match msg {
        WorkletCommand::Drop(id) => {
            let mut global = ctx.env.get_global()?;
            global.delete_named_property(&id.to_string()).unwrap();
        }
        WorkletCommand::Process(args) => {
            process_audio_worklet(ctx.env, args)?;
        }
    }
}
```

orottier:
Okay interesting.

First step is to add loads of logging I guess. 

I like the timeout recv option, will try. But if the underlying issue is a deep sleep before the first process of the quantum, we will still have issues. 

 I will add a check in the AudioWorkletNode constructor to await the worker being ready to collect process commands. That might solve the not-running issue.

I'm travelling today so will be on and offline infrequently

b-ma:
> when I run the examples/audio-worklet.mjs now it only starts properly 10-25% of the time and stays stuck at beginning the rest the time

This one seems to be fixed w/ [5a31ffa](https://github.com/ircam-ismm/node-web-audio-api/pull/124/commits/5a31ffa3ad1a471ad3de7c5fd05661654ec5d539) which makes sens, if the JS worklet was not ready, the `tail_time` was never sent to back to rust which just blocked

For the glitches, this is strange because on my side for example with `audio-worklet-shared-array-buffer`, it seems to be
- 1 sec of clean sawtooth
- 1 sec of clean square
- 1 sec of glitches, and always 3 of them

Harder to tell with the bitcrusher one but seems to be the same pattern, just as if the glitch appends when we are closing the context

b-ma:
> just as if the glitch happens when we are closing the context

Seems confirmed on my side, but I don't really understand what happens there, some random ideas:
- GLOBAL_PROCESS_CALL seems to be never cleared?
- What happens if `audioContext.close()` is called while the worker waits back for the `tail_time` event?

b-ma:
Ok, looks like the process properly exit now w/ [5f75ab5](https://github.com/ircam-ismm/node-web-audio-api/pull/124/commits/5f75ab508e80322acd19f8feca9d1032972948d5).

Not sure however if this is the best way to go, so feel free to butcher it if needed :) 
I tried to be clear why I made things in the comments, but let me know if something still needs further explanations

b-ma:
Ok I think I may have found the way of not allocating the `Float32Array` on each call (at least if number 

```rs
#[macro_use]
extern crate napi_derive;

use napi::*;

#[allow(clippy::needless_lifetimes)]
pub(crate) fn to_byte_slice<'a>(floats: &'a [f32]) -> &'a [u8] {
    unsafe { std::slice::from_raw_parts(floats.as_ptr() as *const _, floats.len() * 4) }
}

#[allow(clippy::needless_lifetimes)]
pub(crate) fn to_float_slice<'a>(bytes: &'a [u8]) -> &'a [f32] {
    unsafe { std::slice::from_raw_parts(bytes.as_ptr() as *const _, bytes.len() / 4) }
}

#[module_exports]
fn init(mut exports: JsObject) -> Result<()> {
  // ...
}

#[js_function]
fn create_typed_array(ctx: CallContext) -> Result<JsTypedArray> {
  let length = 128;
  let array_buffer = ctx.env.create_arraybuffer(length * 4)?;
  let float32_arr = array_buffer
    .into_raw()
    .into_typedarray(napi::TypedArrayType::Float32, length, 0)
    .unwrap();

  Ok(float32_arr)
}

#[js_function(1)]
fn run_process(ctx: CallContext) -> Result<JsUndefined> {
  // this represent our input/output buffers
  let mut value = [0.; 128];
  // populate with dummy values
  for i in 0..128 {
    value[i] = i as f32;
  }

  // get our Float32Array buffer
  let float32_arr = ctx.get::<JsTypedArray>(0)?;
  // ArrayBufferValue seems to be some kind of facade over [u8]
  // https://docs.rs/napi/latest/napi/struct.JsArrayBufferValue.html
  let mut array_buffer_value = float32_arr.into_value()?.arraybuffer.into_value()?;
  // copy input into buffer
  let slice = to_byte_slice(&value);
  array_buffer_value.copy_from_slice(slice);

  // grab process method
  let global = ctx.env.get_global()?;
  let process = global.get_named_property::<JsFunction>("process")?;
  // Get float32array back from args to have owned value to call apply
  // This will probably not be needed in worlet, as arguments are Arrays we don't touch
  let float32_arr = ctx.get::<JsTypedArray>(0)?;
  // execute worklet process
  let _ret: JsUnknown = process.apply1(global, float32_arr)?;

  // copy back ArrayBuffer into `value`
  let float32_arr = ctx.get::<JsTypedArray>(0)?;
  let array_buffer_value = float32_arr.into_value()?.arraybuffer.into_value()?;
  let slice = to_float_slice(&array_buffer_value);
  value.copy_from_slice(slice);

  println!("value {:?}", &value[..]);
  // [1, 1, 1]

  ctx.env.get_undefined()
}
```

With related JS code

```js
const native = require('./test-typedarray.darwin-arm64.node');

globalThis.process = function(output) {
  console.log('hello', output); // [1, 2, 3, ...]

  for (let i = 0; i < 128; i++) {
    output[i] = 1; // fill with 1
  }
}

const array = native.createTypedArray();
console.log(array); // [0, 0, 0, ...]

native.runProcess(array);
console.log(array); // [1, 1, 1, ...]
```

I will have a try later

b-ma:
And finally some wpt results (with lots of timeouts :) !

```
  RESULTS:
  - # pass: 215
  - # fail: 43
  - # type error issues: 0
> wpt duration: 2:34.750 (m:ss.mmm)
```

src/audio_worklet_node.rs
```diff
@@ -51,6 +51,8 @@ struct ProcessorArguments {
 struct ProcessCallChannel {
     send: Sender<WorkletCommand>,
     recv: Receiver<WorkletCommand>,
+    // mark that the worklet has been exited to prevent any further `process` call
```
orottier, line 54: Cool to have a clean exit now. I will indeed try to butcher this a bit, I think this will be better modeled with a new type of WorkletCommand instead, to solve any `recv()` call hanging

orottier (commented):
Hey, good stuff. I'm travelling a few days so had not had time to write any code - also the following days probably.
The WPT results are not bad. I think an issue in rendering OfflineAudioContext is that we should block progress before the worker is online.
Somehow the underrun ratio has improved a bit on my machine, but every few runs of `audio-worklet.mjs` there still occurs one..
Anyway, hope to contribute a bit later!

examples/worklets/bitcrusher.js
```diff
@@ -0,0 +1,77 @@
...
+            minValue: 0,
+            maxValue: 1
+        }];
+    }
+
+    constructor(options) {
+      console.log(`++ in constructor: ${JSON.stringify(options, null, 2)}\n`);
+      // The initial parameter value can be set by passing |options|
+      // to the processor's constructor.
+      super();
```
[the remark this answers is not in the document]
b-ma: Hum my fault... I was abusing `options` to pass all sorts of things (`port`, etc), but then calling `super(option)` was mandatory which is wrong, so I have been to far in the other side... Anyway this should should still work I think, I will check

b-ma:
> The WPT results are not bad

I just can't let you say that this way, they are just amazing, they just exists! (I'm kind of really happy we have numbers and not just a crashed and/or stuck process :)

examples/worklets/bitcrusher.js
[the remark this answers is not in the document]
b-ma: Actually no, this seems to be an issue with the bitcrusher example from spec, the AudioWorkletNodeProcessor itself has no argument, cf. https://webaudio.github.io/web-audio-api/#AudioWorkletProcessor

b-ma:
> Ok I think I may have found the way of not creating the Float32Arrays on each call

I had a try with parameters in [7264803](https://github.com/ircam-ismm/node-web-audio-api/pull/124/commits/726480372c6ec2514621ddfa04ef7987f70eb5ec), this seems to work quite well!
It will be a bit more tricky for inputs / outputs but I'm more and more convinced this is the way to go, let me know what you think

examples/worklets/bitcrusher.js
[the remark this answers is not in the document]
b-ma: Just checked in browsers, and passing or not the options object to `super` has absolutely no impact on initial parameter values, so I confirm this comment seems to be wrong.

However, I realized that the whole option object is passed to the `AudioWorkletNode`, not just the `processorOptions`, will fix that.

b-ma:
For remaining `avoid allocations in process call and safety` issues, I think the following algorithm would do the trick

1. Check `inputs` layout `[n][m]` compared to `js_inputs` (only `m` can change at runtime, as number of inputs is defined at construction)
2. If layout is not the same we need to rebuild `js_inputs` (_allocations only happen at this step_)
    * create new `js_inputs` Array
    * for each input 
      + create new `js_input` Array
      + copy existing `Float32Array` channels into new `js_input` 
      + adjust missing or extra channels
      + freeze `js_input` Array
    * freeze `js_inputs` Array
    * re-assign `js_inputs` to processor
3. Repeat 1. and eventually 2. with `outputs` and `js_ouputs`
4. Copy `inputs` buffers into `js_inputs` using `copy_from_slice`
5. Execute `js_process`
6. Copy `js_ouputs` buffers into `outputs` using `copy_from_slice`

orottier:
Hey, I skimmed your commits and they look good. Nice to see progress on allocation free JS because it may be the garbage collector playing us parts in realtime safety.
Indeed very nice results on the WPT.

I have been benchmarking a bit and it's surprising to see the click actually happen inside the `run_audio_worklet_global_scope` call, not outside. This is nice because apparently we are not fighting the event loop but something inside our call into the worklets (still it may be a garbage collect)..

In any case, if I change 
```
-    while let Ok(msg) =
-        process_call_receiver(worklet_id).recv_timeout(std::time::Duration::from_micros(100))
-    {
+    while let Ok(msg) = process_call_receiver(worklet_id).try_recv() {
```

The load goes up from 11% CPU to 14% on my machine (because we removed some idling) but all the clicks/pop/underruns also seem to dissappear. Can you confirm?

src/audio_worklet_node.rs
```diff
@@ -188,44 +247,46 @@ fn process_audio_worklet(env: &Env, processors: &JsObject, args: ProcessorArgume
...
+                processor.set_property(k_worklet_inputs, new_js_inputs)?;
+                // Override js_inputs with new reference
+                js_inputs = processor.get_property::<JsSymbol, JsObject>(k_worklet_inputs)?;
             }
 
-            let js_outputs = processor.get_property::<JsSymbol, JsObject>(k_worklet_outputs)?;
-
-            for (output_number, output) in outputs.iter().enumerate() {
-                let mut channels = js_outputs.get_element::<JsObject>(output_number as u32)?;
+            if !check_same_io_layout(&js_outputs, outputs) {
```
orottier, line 269: I was in doubt if it was possible for output channel counts to change at runtime, but it is indeed possible:
https://webaudio.github.io/web-audio-api/#configuring-channels-with-audioworkletnodeoptions (note at item 4)

src/audio_worklet_node.rs
```diff
@@ -247,6 +308,31 @@ fn process_audio_worklet(env: &Env, processors: &JsObject, args: ProcessorArgume
...
+                        for (channel_number, channel) in output.iter().enumerate() {
+                            let js_channel =
+                                js_output.get_element::<JsTypedArray>(channel_number as u32)?;
+                            let js_channel_value = js_channel.into_value()?;
+                            let js_channel_buffer: &[f32] = js_channel_value.as_ref();
+
+                            let src = js_channel_buffer.as_ptr();
+                            let dst = channel.as_ptr() as *mut f32;
+
+                            unsafe {
```
orottier, line 332: We can just safely use `copy_from_slice`, no?

src/audio_worklet_node.rs
```diff
@@ -188,44 +247,46 @@ fn process_audio_worklet(env: &Env, processors: &JsObject, args: ProcessorArgume
...
+            // Copy inputs into JS inputs buffers
+            for (input_number, input) in inputs.iter().enumerate() {
+                let js_input = js_inputs.get_element::<JsObject>(input_number as u32)?;
 
-                // delete remaining channels, if any
-                for i in output.len() as u32..channels.get_array_length().unwrap() {
-                    channels.delete_element(i)?;
+                for (channel_number, channel) in input.iter().enumerate() {
+                    let js_channel = js_input.get_element::<JsTypedArray>(channel_number as u32)?;
+                    let mut js_channel_value = js_channel.into_value()?;
```
orottier, line 284: Would be good to check at some point if we don't allocate with some of these calls

orottier (commented):
The code looks good to me, I have left some nitpicks.

However, the garbage collector stats before this change are equal to afterwards:

`node --trace-gc examples/audio-worklet.mjs`
gives e.g.
`[46986:0x130028000]     5658 ms: Scavenge 5.9 (9.0) -> 4.3 (9.0) MB, 0.17 / 0.00 ms  (average mu = 1.000, current mu = 1.000) task;`
Meaning it reclaimed from 5.9 to 4.3 MB.

After the change:
`[47391:0x128028000]     5354 ms: Scavenge 5.9 (8.7) -> 4.3 (8.7) MB, 0.12 / 0.00 ms  (average mu = 1.000, current mu = 1.000) task;`

Not sure what to make from this..

orottier:
The GC pauses appear to be of the range 0.08 - 0.17 ms and seem not related to the excessive call durations I'm measuring.. So what is stalling my audio thread!?

src/audio_worklet_node.rs
b-ma: Yup indeed, I think there might be an issue on the rust side on this point, the `audioworkletnode-channel-count.https.html` wpt test is a good example of what is expected. The test is currently crashing for some reason I didn't really understand, but it seems the rs output length keeps to be 1 even when input is 3

src/audio_worklet_node.rs
b-ma: That was my first try but I did manage to find a way to `iter_mut()` the `&'static [&'static [&'static [f32]]]` output...

src/audio_worklet_node.rs
b-ma: Normally except in `rebuild_io_layout` there should be no allocation, but clearly needs to be confirmed

I'm just thinking we could also have a pool of preallocated Float32Array to mitigate further this possible issue

orottier:
Okay after loads of logging I have come to the surprising conclusion that the call `.recv_timeout(std::time::Duration::from_micros(100))` is the culprit. It takes up to 8 milliseconds sometimes which is pretty insane to think about! Reading the issue list of crossbeam seems to indicate that they are doing some weird stuff with spinning and backoffs and possibly that does not work nice on ARM? In any case this is not suitable for our latency-sensitive ping pong of the audio thread and the worker.

So as I already mentioned changing the call to just `try_recv` solves my underrun issues. I did a bit of further profiling because I was worried the worker thread would max out a single CPU core but I don't see it happening, so possible the event loop does a bit of idling. Which is probably fine for now. The load per core does not exceed 15% for audio-worklet.mjs

I definitely want to look into this further, possible using some primitives from the `parking_lot` crate to perform the ping pong with CondVars maybe. But I guess that is not a must have before merging.

src/audio_worklet_node.rs
orottier: Looking better into this, I think your solution is just good as it is.
We can't use the WorkletCommand because it is strictly from rust audio thread to js worker, and here we need to move information vice versa. I don't see a way to further simplify this.

orottier:
Regarding 
> TODO: In the case of OfflineAudioContext, it is probably desired to block here

We could maybe add a method `await_all_processors_registered` in `js/AudioWorklet.js` to ensure the number of `create-processor` is equal to the number of `processor-registered` and call it before the OfflineAudioContext startRendering?

b-ma:
Cool, nice that you found the issue! I think we could start to consider merging this, and put all remaining ideas / improvements to a new issue. 

Is there anything you think we should do really before?

orottier:
Yeah sounds good! Perhaps only my final comment

b-ma:
Yup sure, I will have a look

b-ma:
Follow up in #127

b-ma:
\o/ great work!

jampekka:
Awesome work! :fire:

orottier:
Quite the milestone indeed :) Congrats to us. And also thanks @jampekka for chiming in on the exploratory phase. An extra pair of eyes is always very motivating.

orottier:
Please let me know when you have released this on npm. I think it is worth the announcement on the web audio slack. Also I would like to get some tips and tricks from Paul and Hoch in the WG afterwards

b-ma:
Done [node-web-audio-api@0.21.0](https://www.npmjs.com/package/node-web-audio-api) !

> Also I would like to get some tips and tricks from Paul and Hoch in the WG afterwards

Yup, I will try to attend too next times

PR #128, opened 2024-05-27
Yield once to VM per quantum, another thread prio crate, some timings, use condvar instead of crossbeam; no improvements

orottier:
Not really sure what to make out of all this. My CPU load goes down by a factor 2 but now my real-time issues are back.

Leaving this here for further consideration, my advices would be:
- keep padenot's thread prio crate, it looks solid
- keep the condvar, more lightweight than crossbeam and also some code improvements
- remove the tick processor and still yield to V8 continuously (old behaviour) - somehow my mac seems to go in deep sleep if the worker thread has nothing to do and may wake up only milliseconds later?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,6 @@ audio_thread_priority = "0.34.0"
 crossbeam-channel = "0.5"
 napi = { version="2.16", features=["napi9", "tokio_rt"] }
 napi-derive = { version="2.16" }
-thread-priority = "1.2"
 web-audio-api = "1.2"
 # web-audio-api = { path = "../web-audio-api-rs" }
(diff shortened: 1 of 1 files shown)


b-ma:
Hey, thanks

I will need to have a closer look (and play a bit with CondVar in a more simple setup I guess :) to understand your changes more clearly.

> now my real-time issues are back....

Weird cannot reproduce on my side, I tried to have a quick look to libuv issues and docs but couldn't find any mention of some dynamic priority management of the threads, so the mystery still holds...

In any case, I agree with your recommendations, having no clicks is far more important than having this one per quantum tick! This can clearly wait for clean solution

Issue #118, opened 2024-04-20
[priority] Memory leak introduced after v0.18


b-ma:
Nodes are not garbage collected as they should, clearly visible with granular scrub example...

Do not use v0.19

b-ma:
This seems to be the faulty commit: https://github.com/ircam-ismm/node-web-audio-api/commit/8d819d49f8fae662b73a1059d0054ab36a4904da

b-ma:
still problematic on RPi in 0.21.0 (only half the problem seems to be solved...)

b-ma:
- [ ] bisect between 0.18 and 0.19 to hopefully find another faulty commit
- [ ] AudioBuffer could be refactored to raw object maybe same issue as for AudioParam (this was there before 0.18 but not wrapped in a facade...)

b-ma:
First step is to cross-compile `node-web-audio-api.linux-arm64-gnu.node` target in a docker image as building on the RPi itself take dozens of minutes, which makes the thing impossible to debug and test in practice

orottier:
Do you see the memory leak only in the RPi? We would expect it to be present on all platforms, no?

You can use the hidden method `AudioContext::run_diagnostics` to print the current audio graph to see if nodes are not reclaimed properly. The issue might also be somewhere else of course (unhandled events, etc)

b-ma:
> Do you see the memory leak only in the RPi? We would expect it to be present on all platforms, no?

Yup you are right, it's just that it creates really audible cracks on RPi (with CPU going up to 50% instead of 10-15%) while it sounds ok on my machine. But after rechecking I can indeed see the CPU going from ~2.6 to almost 9% too... while with version 0.18 it does not go beyond 3%

> You can use the hidden method AudioContext::run_diagnostics to print the current audio graph to see if nodes are not reclaimed properly. The issue might also be somewhere else of course (unhandled events, etc)

Thank for the hint I will have a look. My guess is that this to is due to something in how the bindings are made which prevent GC and makes the graph grow and grow. My best suspects right now are:
- Internally created classes wrapped in facade and then used as attributes, cf. previous issue with `AudioParam` which seemed to create the exact same problem
- The way event callbacks are handled (maybe the first simple to test as a first step)

I'm a bit nervous with that stuff considering all the work and refactoring that has been made since version 0.18...

b-ma:
[...]

PR #130, opened 2024-06-05
Fix issue with garbage collection

b-ma:
cf. #118 

Ok this is the `ended` event logic which produced the issue (I will recheck but I don't think this is critical for other events)

For the record:

> AudioScheduledSourceNode are not collected in Scavenge but only in Mark-Compact step, while GainNodes are:
> 
> ```
> [20747:0x130008000]     6339 ms: Scavenge 9.2 (12.7) -> 8.7 (17.7) MB, pooled: 0 MB, 0.96 / 0.04 ms  (average mu = 1.000, current mu = 1.000) task;
> NAPI: NapiGainNode dropped
> NAPI: NapiGainNode dropped
> // ....
> [20747:0x130008000]     8119 ms: Mark-Compact (reduce) 10.1 (17.7) -> 5.0 (10.0) MB, pooled: 0 MB, 3.04 / 1.00 ms  (+ 1.4 ms in 0 steps since start of marking, biggest step 0.0 ms, walltime since start of marking 6 ms) (average mu = 0.999, current mu = 0.999) finalize incremental marking via task; GC in old space requested
> NAPI: NapiAudioBufferSourceNode dropped
> NAPI: NapiAudioBufferSourceNode dropped
> NAPI: NapiAudioBufferSourceNode dropped
> ```

Looking at CPU usage with htop (ok it is what it is :), we can even see the performance gain from upstream crate I think
- v0.18 - 2.4 / 3.8% CPU
- here   - 1.9 / 3.0% CPU

Now the question is, can we fix this somehow or should we drop these events until we find a clean solution....

Next steps:
- [ ] hotfix release which drops `ended` events
- [ ] try to fix the problem

### DIFF
--- a/js/AudioScheduledSourceNode.js
+++ b/js/AudioScheduledSourceNode.js
@@ -33,11 +33,12 @@ class AudioScheduledSourceNode extends AudioNode {
 
     // Add function to Napi object to bridge from Rust events to JS EventTarget
     // It will be effectively registered on rust side when `start` is called
+    //
+    // Note 2024-06-05 - We use bind instead of arrow function because arrow function
+    // prevent the node to be collected by Scavenge step of GC, which can lead to
+    // oversized graphs and performance issues.
+    // cf. https://github.com/ircam-ismm/node-web-audio-api/tree/fix/118
     this[kNapiObj][kOnEnded] = (function(_err, rawEvent) {
-      if (typeof rawEvent !== 'object' && !('type' in rawEvent)) {
-        throw new TypeError('Invalid [kOnEnded] Invocation: rawEvent should have a type property');
-      }
-
       const event = new Event(rawEvent.type);
       propagateEvent(this, event);
     }).bind(this);
--- a/js/AudioScheduledSourceNode.js
+++ b/js/AudioScheduledSourceNode.js
@@ -33,14 +33,14 @@ class AudioScheduledSourceNode extends AudioNode {
 
     // Add function to Napi object to bridge from Rust events to JS EventTarget
     // It will be effectively registered on rust side when `start` is called
-    // this[kNapiObj][kOnEnded] = (err, rawEvent) => {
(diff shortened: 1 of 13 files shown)


orottier:
> AudioScheduledSourceNode are not collected in Scavenge but only in Mark-Compact step, while GainNodes are:

This could indicate a circular reference, e.g. an object references a closure while the closure also references the object. This is harder to garbage collect.

I'm fine with a hotfix breaking the ended event if that helps your RPi issues. Does it actually? Have you tested this on RPi?

js/AudioScheduledSourceNode.js
```diff
@@ -33,14 +33,14 @@ class AudioScheduledSourceNode extends AudioNode {
...
-    //   propagateEvent(this, event);
-    // };
+    this[kNapiObj][kOnEnded] = (function(_err, rawEvent) {
+      if (typeof rawEvent !== 'object' && !('type' in rawEvent)) {
+        throw new TypeError('Invalid [kOnEnded] Invocation: rawEvent should have a type property');
+      }
+
+      const event = new Event(rawEvent.type);
+      propagateEvent(this, event);
+    }).bind(this);
```
b-ma, line 43: Thanks for the hint, using explicit bind instead of arrow function seems to fix the issue

b-ma:
> Does it actually? Have you tested this on RPi?

Not yet, I will try to find some to do that today but I'm pretty confident, the cracks and their end seemed related to `Mark-Compact` GC calls

js/AudioScheduledSourceNode.js
b-ma: Really surprised such a little thing caused so much problems actually, I was always thinking arrow functions were some kind of syntactic sugar for `bind`...

b-ma:
Ok seems to work on RPi too! I will generalize the pattern to all events and make a release on Friday

orottier:
Super!

PR #131, opened 2024-06-07
Fix wpt

b-ma:
This fixes some issues that were introduced by AudioWorklet (cf. `getElementByTagName` spam, 100% CPU, etc.)

PR #132, opened 2024-06-07
Feat - Use buffer pool in `AudioWorkletGlobalScope`

b-ma:
Built on top of #131 

Use a thread pool in worklets to minimize allocations and ensure that input and output cannot be transfered to main thread

```
  RESULTS:
  - # pass: 7299
  - # fail: 527
  - # type error issues: 3
> wpt duration: 2:50.251 (m:ss.mmm)
```

### DIFF
--- a/js/AudioWorkletGlobalScope.js
+++ b/js/AudioWorkletGlobalScope.js
@@ -315,13 +315,23 @@ parentPort.on('message', event => {
     case 'node-web-audio-api:worklet:add-module': {
       const { code, promiseId } = event;
       const func = new Function('AudioWorkletProcessor', 'registerProcessor', code);
-      func(AudioWorkletProcessor, registerProcessor);
 
-      // send registered param descriptors on main thread and resolve Promise
-      parentPort.postMessage({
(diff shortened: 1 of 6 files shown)


js/AudioWorkletGlobalScope.js
```diff
@@ -21,14 +23,72 @@ const kWorkletInputs = Symbol.for('node-web-audio-api:worklet-inputs');
...
+
+  get() {
+    if (this.#pool.length === 0) {
+      return this.#allocate();
+    }
+
+    return this.#pool.pop();
+  }
+
+  recycle(buffer) {
```
orottier, line 71: We don't need to set the values to zero because on each render call we copy all input values anyway?

orottier (approved):
Alright, it's quite a bit of extra code (or framed differently: attack surface for bugs) but I guess if it helps real-time safety we should do it.
Looks good to me.

js/AudioWorkletGlobalScope.js
b-ma: Yup!

PR #136, opened 2024-09-06
Fix #133

### LINKED ISSUE #133: OfflineAudioContext: reuse or clean memory resources
Hi!

When performing consecutive renders with the OfflineAudioContext, memory usage only accumulates, for example with repeated calls like below (ran with the `--expose-gc` flag).

It might seem like an approach to reuse the OfflineAudioContext instance (e.g. by instantiating it outside the loop), but that results in an error after the first iteration. Uncommenting the suspend / resume calls also results in an error after a few iterations.

Do you see any options to clean up the used resources in this example? - or would some modifications to the library be required?

```
import { OfflineAudioContext } from 'node-web-audio-api';

for( let i=0; i < 100000; i++ ) {
  console.log('+ i:', i);

  let offline = new OfflineAudioContext(1, 10*48000, 48000);
  
  // offline.suspend(128 / 48000).then(async () => {
    const osc = offline.createOscillator();
    osc.connect(offline.destination);
    osc.frequency.value = 220;
    osc.start(0.);
    osc.stop(10.);
  
  //   await offline.resume();
  // });
  
  const buffer = await offline.startRendering();
  console.log('+ buffer duration:', buffer.duration);

  offline = null;

  if (global.gc) {
    global.gc();
  }
}
```

Running this on Apple M1 Pro

b-ma:


generator/rs/lib.tmpl.rs
```diff
@@ -1,4 +1,6 @@
 #![deny(clippy::all)]
+// @todo - properly fix this clippy issue
+#![allow(clippy::zero_repeat_side_effects)]
```
orottier, line 3: meh, I think this is a napi issue causing false positive in clippy

orottier (approved):
Looks good!

js/OfflineAudioContext.js
```diff
@@ -82,24 +82,26 @@ module.exports = function patchOfflineAudioContext(jsExport, nativeBinding) {
...
-        }
+        this.#renderedBuffer = new jsExport.AudioBuffer({ [kNapiObj]: rawEvent.renderedBuffer });
 
         const event = new jsExport.OfflineAudioCompletionEvent(rawEvent.type, {
           renderedBuffer: this.#renderedBuffer,
         });
 
-        propagateEvent(this, event);
+        // delay event propagation to next tick that it is executed after startRendering fulfills
+        setTimeout(() => {
```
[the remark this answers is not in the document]
b-ma: Yup thanks, sure I have been misleaded by mdn which marks it deprecated but it doesn't stand in node context

generator/rs/lib.tmpl.rs
b-ma: Yup I agreed, let's see if the problem disappear in a future napi-rs release, I opened an issue to not forget that 
cf. https://github.com/ircam-ismm/node-web-audio-api/issues/137

PR #152, opened 2025-03-05
Add missing types

mikezaby:
Hello, I make this PR to add the missing types, also I re-arrange them so they have almost the same order with `index.mjs`


b-ma:
Thanks! I will release this soon

PR #153, opened 2025-04-11
Add Caveat about DeviceNotAvailable

qwelias:


qwelias:
I get caveats is probably not the best place since it can happen on the web as well, but this is a much more likely scenario on the backend

b-ma:
Thanks!

PR #162, opened 2025-11-25
Add Faust examples

sletz:
This PR adds three examples using the [faustwasm](https://www.npmjs.com/package/@grame/faustwasm) package, dynamically compiling and then running [Faust](https://faust.grame.fr) DSP programs. The code currently uses the `FaustMonoDspGenerator` to create synthesisers or effects. 

If desired, the `FaustPolyDspGenerator` generator could possibly be demonstrated later on to create MIDI controllable instruments.


package.json
orottier: Great to see this is working, but we should probably not add Faust to the dependencies. Maybe have it in the devDependencies?

sletz: Done is this new PR: https://github.com/ircam-ismm/node-web-audio-api/pull/163

PR #164, opened 2026-01-01
faustwasm updated to 0.14.0.

sletz:


b-ma:
Hey thanks! 
Just let me now when it is stabilizing on your side and you want me to merge (and happy new year :)

b-ma:
Thanks!

> I might be mistaken but I don't think updating a devDependency has an effect on end users

No indeed, it doesn't

sletz:
Reading here: https://docs.npmjs.com/specifying-dependencies-and-devdependencies-in-a-package-json-file, and having faustwasm packagage in devDependency, what is expected work flow for user using NPM packaged version when  Faust support is needed ?

b-ma:
Hey @sletz ,

I just tried to update your faust examples using the second cleaner syntax given that the lib now exposes a `polyfill.js` entry point. 

The problem is that it fails because the polyfill creates a `window` object to support existing libs that explicitly rely on `window`, e.g. https://github.com/chrisguttandin/standardized-audio-context

I have seen that you check if we run in a browser using `typeof window === 'object'` (e.g. https://github.com/grame-cncm/faustwasm/blob/79248357f2a9e081f84bb4418fcfc70b4b61be83/src/instantiateFaustModuleFromFile.ts#L18), then we fall into the wrong branch. I tried to replace it with a more robust 

```js
const isBrowser = new Function('try {return this===window;}catch(e){ return false;}');
// and similarly
const isNode = new Function('try {return this===global;}catch(e){ return false;}');
```

But then it fails again because of some similar check are done in `libfaust-wasm.js` which appear as transpiled code and I could not figure out where it comes from.

Happy to help (or make a PR with some guidance) if you are interested in fixing this

PR #166, opened 2026-01-16
feat: register global + tone.js example

b-ma:
Minimal example for Tone.js

Several open questions:
- should we include this `register-global.js` thing in the lib, looks like "yes" as it may really help to work with existing libs / frameworks, helpful to use faust (cf. https://github.com/ircam-ismm/node-web-audio-api/pull/164#issuecomment-3758998865) or `wam` plugins for example.
- cannot get rid of `Tone.setContext(audioContext)`, somehow related to how `standardized-audio-context` works but couldn't find the reason

Maybe would be nice to work on the compatibility with `standardized-audio-context` first?

### DIFF
--- a/examples/tone.js
+++ b/examples/tone.js
@@ -1,0 +1,17 @@
+import '../register-global.mjs';
+import * as Tone from 'tone';
+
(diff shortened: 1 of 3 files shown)


b-ma:
Implemented in #181 

Released in v2.0.0

Issue #156, opened 2025-10-28
Upgrade to napi#v3


b-ma:
napi-rs used version: 2.16
napi-rs latest version: 3.8.3

## Requirements

- should not impact wpt results
- should be able to run all existing examples
- keep general architecture with JS facade / Napi wrappers / Native lib
- only Napi wrappers should be impacted

## Roadmap

- [x] sketch strategy for class instances as class members (e.g. AudioContextDestination, AudioParam)
- [x] sketch strategy for connect / disconnect
- [x] implement minimal / rust only code to run `many-oscillators-with-env.js` example
- [x] clean repo to have only a few classes, reintroduce JS facade
- [x] implement `AudioBuffer` and `AudioScheduledSourceNode`
- [x] run `granular-scrub` example
- [x] review `OfflineAudioContext` 
- [x] remaining generated AudioNodes
- [x] events
- [x] render capacity
- [x] AudioWorklet
- [ ] review ScriptProcessor strategy (see if we use the same one as for AudioWorklet and avoir Buffer pack / unpack)
- [ ] review AudioNode options parsing 
- [ ] back to previous WPT results

b-ma:
Working repo: https://github.com/b-ma/node-web-audio-api-napi-3

b-ma:
Now in #172 draft PR

b-ma:
Done in #172

PR #172, opened 2026-03-30
Refactor/napi 3

b-ma:
cf. #156


orottier:
Hey this is looking very good. Anything you would need me for?

b-ma:
Hey thanks, I'm pretty happy with the result so far indeed (this is starting to look like proper Rust :)

For now things are still a bit messy and interleaved, but at some point that would be really helpful to have some help to track WPT regressions, etc. I'll let you know once things start to stabilise

b-ma:
Current wpt results:

```
  RESULTS:
  - # pass: 4476
  - # fail: 641
  - # type error issues: 1
> wpt duration: 1:48.358 (m:ss.mmm)
```

For the record, last reported results on main branch: https://github.com/ircam-ismm/node-web-audio-api/issues/57#issuecomment-4109272272

```
RESULTS:
- # pass: 5061
- # fail: 382
- # type error issues: 0
```

b-ma:
Ok things start to look quite good here

- CI is happy most of the time
- Small tests within repo are passing 
- All examples are running 
- No notable difference in benchmarks

# WPT

For WPT, things get a bit more strange, nothing really alarming but they are still a bit lower than before and this is hard to track down and understand exactly why...

Note that I updated the submodule, I guess this explains the difference in `main` branch with the number reported in my last comment

## Global

```
[main] RESULTS:
  - # pass: 5027
  - # fail: 436
  - # type error issues: 0
> wpt duration: 6:23.340 (m:ss.mmm)

[napi-3] RESULTS:
  - # pass: 5001
  - # fail: 440
  - # type error issues: 0
> wpt duration: 5:59.016 (m:ss.mmm)
```

## Dissect

Only differences are reported, but they don't explain the diff in global numbers, this is a bit weird

### first pass

```
[main] the-audiocontext-interface
~~~~~~~~~~~~~~~~~
  RESULTS:
  - # pass: 58
  - # fail: 41
  - # type error issues: 0

[napi-3] the-audiocontext-interface
~~~~~~~~~~~~~~~~~
  RESULTS:
  - # pass: 57
  - # fail: 43
  - # type error issues: 0
```
```
[main] the-offlineaudiocontext-interface
~~~~~~~~~~~~~~~~~
  RESULTS:
  - # pass: 49
  - # fail: 10
  - # type error issues: 0

[napi-3] the-offlineaudiocontext-interface
~~~~~~~~~~~~~~~~~
  RESULTS:
  - # pass: 52
  - # fail: 7
  - # type error issues: 0
 ```

### second pass

```
the-audiocontext-interface - no difference this time
```
```
[main] the-audioworklet-interface
~~~~~~~~~~~~~~~~~
  RESULTS:
  - # pass: 141
  - # fail: 32
  - # type error issues: 0

[napi-3] the-audioworklet-interface
~~~~~~~~~~~~~~~~~
  RESULTS:
  - # pass: 139
  - # fail: 31
  - # type error issues: 0
```
```
[main] the-dynamicscompressornode-interface
~~~~~~~~~~~~~~~~~
  RESULTS:
  - # pass: 14
  - # fail: 3
  - # type error issues: 0

[napi-3] the-dynamicscompressornode-interface
~~~~~~~~~~~~~~~~~
  RESULTS:
  - # pass: 13
  - # fail: 4
  - # type error issues: 0
```
```
the-offlineaudiocontext-interface - same as first pass (but this is in the right direction)
```

For the the AudioContext, I have already seen small differences between subsequent runs, I suspect weird edge cases / race conditions with suspend / resume, etc. maybe things that also imply the upstream lib

I will continue to work on the `AudioWorklet` a bit more, as they are used in many places in the wpt we need them to be quite solid

**Comment by b-ma** [comment]
WPT results now:

```
  RESULTS:
  - # pass: 5052
  - # fail: 414
  - # type error issues: 0
 ```

Still see some little noise in results (which were also there before refactoring, so this is really hard to track down), but I think the general picture is there.

Ready to merge I think

@orottier, any comment / suggestion on that?

orottier:
Hey, I am actually away this week but i will try to have a look!

orottier:
So it seems to work! Compiled and ran some examples 
 WPT instability is weird indeed, I tried a few things but it is related to our runner setup I believe, not some underlying issue
Have not really looked at the actual code yet

b-ma:
> So it seems to work! Compiled and ran some examples

Cool

> WPT instability is weird indeed, I tried a few things but it is related to our runner setup I believe, not some underlying issue

Seems to be both, e.g. on compressor node reduction is computed while it should not (cannot reproduce in pure rust), but it seems to be more an issue with how we process the graph I think: the node is completely disconnected in the test but `render` is called so reduction is computed against silence while it should stay at its default value

I have also run the granular example for 10 hours on a RPi and it worked like a charm, pretty cool :)

index.d.ts
orottier, line 1: I think this file got trashed somehow. It should not expose any Napi classes? Did we create this file manually before?

package.json
```diff
@@ -55,6 +56,7 @@
     "@ircam/eslint-config": "^2.0.0",
     "@ircam/sc-scheduling": "^1.0.0",
     "@ircam/sc-utils": "^1.9.0",
+    "@napi-rs/cli": "^3.2.0",
```
orottier: Do we need this devDependency?

src/offline_audio_context.rs
```diff
@@ -1,214 +1,86 @@
...
-    pub fn create_js_class(env: &Env) -> Result<JsFunction> {
-        let interface = base_audio_context_interface![
-            Property::new("length")?.with_getter(get_length),
-            Property::new("startRendering")?.with_method(start_rendering),
-            Property::new("resume")?.with_method(resume),
-            Property::new("suspend")?.with_method(suspend)
-        ];
-
-        env.define_class("OfflineAudioContext", constructor, &interface)
+    pub(crate) fn unwrap(&self) -> &OfflineAudioContext {
```
orottier: Nitpick: It was maybe a good time to rename these functions to `inner()`, because `unwrap()` has a little too much meanings now..

src/offline_audio_context.rs
```diff
@@ -1,214 +1,86 @@
...
+    #[napi(getter)]
+    pub fn length(&self) -> u32 {
+        self.inner.length() as u32
+    }
 
-    let length = obj.length() as f64;
-    ctx.env.create_double(length)
-}
+    #[napi(catch_unwind)]
+    pub async fn start_rendering(&self) -> Result<NapiAudioBuffer> {
```
orottier, line 68: The panic message for calling this method twice is not so nice
`[lib/errors.js] Unexpected Rust error [Error: Panic in async function] { code: 'GenericFailure' }`
It is because we unwrap in the base lib in an async method, not much to do about it here I guess. Unless you extend the `throwSanitizedError` to convert "Panic in async function" into some InvalidStateError

src/base_audio_context.rs
```diff
@@ -1,109 +1,115 @@
...
+            pub fn sample_rate(&self) -> f32 {
+                self.inner.sample_rate()
+            }
 
-            ctx.env.create_string(state_str)
-        }
+            #[napi(getter, js_name = "renderQuantumSize")]
+            pub fn render_quantum_size(&self) -> u32 {
+                // self.inner.render_quantum_size(); // @fixme - implement upstream
+                128
```
orottier, line 51: Just merged this https://github.com/orottier/web-audio-api-rs/pull/587 but not released

js/AudioNode.js
```diff
@@ -198,7 +197,6 @@ class AudioNode extends EventTarget {
         output = 0;
       }
 
-      // Rust does not make difference between AudioNode and AudioParam
```
orottier, line 201: Indeed, but you do so this is a great moment to return `undefined` conform the spec when connecting to an AudioParam (instead of a Node)

src/audio_node.rs
```diff
@@ -17,802 +17,610 @@
...
+        impl $napi_struct {
+            #[napi(getter, js_name = "numberOfInputs")]
+            pub fn get_number_of_inputs(&self) -> u32 {
+                self.inner.number_of_inputs() as u32
+            }
 
-            ctx.env.create_string(value_str)
-        }
+            #[napi(getter, js_name = "numberOfOutputs")]
+            pub fn set_number_of_outputs(&self) -> u32 {
```
orottier:
```suggestion
            pub fn get_number_of_outputs(&self) -> u32 {
```

src/audio_node.rs
```diff
@@ -17,802 +17,610 @@
...
+                self.inner
+                    .set_channel_interpretation(channel_interpretation);
+            }
 
-                match arg0.get_type()? {
-                    ValueType::Number => {
-                        let output = arg0.coerce_to_number()?.get_double()? as usize;
-                        native_src.disconnect_output(output);
+            #[napi(catch_unwind)]
+            pub fn connect(
```
orottier, line 89: This function is.. interesting, but I see what is happening. The previous version is not so nice either..

js/AudioParam.js
orottier, line 92: Not related to the PR, but we shouldn't do silent validation and return here no? Just throw a rust error when mapping to an AutomationRate enum value?

src/audio_param.rs
```diff
@@ -1,229 +1,106 @@
...
+    #[napi(getter, js_name = "value")]
+    pub fn get_value(&self) -> f32 {
+        self.inner.value()
+    }
 
-    let value = ctx.get::<JsObject>(0)?.coerce_to_number()?.get_double()? as f32;
-    let start_time = ctx.get::<JsObject>(1)?.coerce_to_number()?.get_double()?;
-    let time_constant = ctx.get::<JsObject>(2)?.coerce_to_number()?.get_double()?;
-    obj.set_target_at_time(value, start_time, time_constant);
+    #[napi(setter, js_name = "value")]
```
orottier: Should also catch_unwind? Or does this never panic on the rust side? I can't recall..

orottier (commented):
Okay I can't say I fully read and understood everything, but I think moving to napi3 with the new style wrapping is a pretty nice improvements. And if we have test and performance parity with the previous version, that is very nice!!

orottier:
I have not really looked at any worklet related files btw, if there are structural changes in there, those need some scrunity too.. But not today!

index.d.ts
b-ma: Yup it has been created manually, I trashed it when playing around with `napi`, nice catch thanks!

package.json
b-ma: Seems I managed to remove it completely :)

src/offline_audio_context.rs
b-ma: done

js/AudioNode.js
b-ma: Thanks, never realized that, done

src/audio_param.rs
b-ma: Nice catch I just forgot to add it, this can indeed throw

```
const gain = new GainNode(audioContext);
gain.gain.value = 'hello';
> TypeError: Failed to set the 'value' property on 'AudioParam': The provided float value is non-finite.
```

js/AudioParam.js
b-ma: I can't find it in the spec, but tests in browsers seems to show that setting an enum like / dict type with a wrong value is just ignored without throwing. 

(just retested with AudioParam::automationRate, OscillatorNode::type, looks consistent)

src/audio_node.rs
b-ma: Yup... and this is really clean compared to the previous implementation :)

src/offline_audio_context.rs
b-ma: I have made up something which is not completely satisfying but a bit more useful 

There seems to have something In napi for async panic handling, but somehow we fall in the default branch:
cf.  https://github.com/napi-rs/napi-rs/pull/2050/changes
```rs
  {
    let jh = spawn(inner);
    spawn(async move {
      if let Err(err) = jh.await {
        if let Ok(reason) = err.try_into_panic() {
          if let Some(s) = reason.downcast_ref::<&str>() {
            deferred_for_panic.reject(Error::new(crate::Status::GenericFailure, s));
          } else {
            deferred_for_panic.reject(Error::new(
              crate::Status::GenericFailure,
              "Panic in async function",
            ));
          }
        }
      }
    });
  }
```

Maybe it tells you something?

js/AudioParam.js
orottier: Alright then

orottier:
Thanks for the additional commits. Will have a look at napi/async and the worklet stuff later this week if that is okay

b-ma:
Sure no hurries, thanks !

js/AudioWorklet.js
```diff
@@ -116,72 +117,97 @@ class AudioWorklet {
...
+            break;
+          }
+          case 'node-web-audio-api:worklet:add-module-failed': {
+            const { promiseId, err } = event;
+            const { reject } = this.#idPromiseMap.get(promiseId);
+            this.#idPromiseMap.delete(promiseId);
+            reject(err);
+            break;
+          }
+          case 'node-web-audio-api:worlet:processor-registered': {
```
orottier:
```suggestion
          case 'node-web-audio-api:worklet:processor-registered': {
```
And at other places of course

js/AudioWorkletGlobalScope.js
```diff
@@ -150,30 +158,46 @@ globalThis.AudioWorkletProcessor = class AudioWorkletProcessor {
...
+    this.#errorPort.postMessage({ cmd, err });
   }
 };
 
-// follow algorithm from:
-// https://webaudio.github.io/web-audio-api/#dom-audioworkletglobalscope-registerprocessor
+// Algorithm: https://webaudio.github.io/web-audio-api/#dom-audioworkletglobalscope-registerprocessor
 globalThis.registerProcessor = function registerProcessor(name, processorCtor) {
   const parsedName = conversions['DOMString'](name, {
     context: `Cannot execute 'registerProcessor' in 'AudoWorkletGlobalScope': name (${name})`,
```
orottier:
```suggestion
    context: `Cannot execute 'registerProcessor' in 'AudioWorkletGlobalScope': name (${name})`,
```

js/AudioWorkletNode.js
orottier, line 1: In `tests/AudioWorklet.spec.mjs` we check
`it('should throw a clean error when processor constructor is invalid', ...` for online AudioContext.
In OfflineAudioContext, the process will hang indefinitely

```
  it('OfflineAudioContext.startRendering should return when processor constructor is invalid', async () => {
    const audioContext = new OfflineAudioContext(1, 128, 48000);
    await audioContext.audioWorklet.addModule('./worklets/invalid-ctor.worklet.mjs');

    const invalid = new AudioWorkletNode(audioContext, 'invalid-ctor');
    invalid.addEventListener('processorerror', e => prettyPrintErr(e.error));

    await audioContext.startRendering();
  });
```

orottier (approved):
Hey, all in all the worklet stuff is also an improvement with napi3 I believe. Nice work

js/AudioWorkletNode.js
b-ma: Nice catch! This was hiding multiple issues :)

The solution is not completely satisfying, but it works let's keep possible improvements for future PRs

b-ma:
If ok for you I think I will merge (I start to add/change things not really related to the initial PR... :)

Still need to define if we consider this a major version or not (technically the API surface didn't change but still a big change...), any opinion on that?

Issue #26, opened 2023-07-23
Simplify running examples


b-ma:
Currently running example requires to build the binary, which implies installing rust toolchain, etc.

Should have a more simple / accessible way of doing that, i.e. relying only on `npm`, while not spoiling dev experience

PR #177, opened 2026-05-21
Refactor - Drop commonjs syntax

### LINKED ISSUE #175: Chore - Drop commonjs support
We can now drop the `.cjs` entry point which will be more standard compliant, and simplify the codebase

This should not impact users as we can `require` esm module since Node v22 (minimum node version for the lib)

cf. https://nodejs.org/api/modules.html#loading-ecmascript-modules-using-require

b-ma:
Fixes #175 (...and #26)

Kind of the JS equivalent of #156 , should not have visible impact for users but cleans many messy workarounds that implies lots of structural changes


b-ma:
```
  RESULTS:
  - # pass: 5108
  - # fail: 471
  - # type error issues: 0
> wpt duration: 3:10.975 (m:ss.mmm)
```

Improvements due in most part to the wpt submodule update

PR #185, opened 2026-05-27
Some AudioWorklet refactoring and improvements

b-ma:
WPT Results are now

**before**
```
- # pass: 168
- # fail: 21
- # type error issues: 0
```

**after**
```
- # pass: 173
- # fail: 15
- # type error issues: 0
```

**Remaining issues:**

- AudioParam does not behave as expected -> needs to be reviewed in upstream
- Configurable render quantum size -> to be implemented in upstream
- Issue with how `currentTime` and `currentFrame` are inited -> a bit quite tricky probably, means we should be able to request the information to the Rust audio thread somehow
- Blob can be deserialized in Node, which makes `audioworkletnode-onerror.https.html` fail
- Transfer of input / output underlying arrayBuffers is forbidden on our side (causes Segmentation Fault) -> really don't see why it is allowed anyway, probably won't fix
- One JS tick per quantum, cf. #128
- Plus one or two minor things


b-ma:
WPT against upstream main 

```
RESULTS:
- # pass: 5137
- # fail: 440
- # type error issues: 0
> wpt duration: 2:44.291 (m:ss.mmm)
```

Issue #159, opened 2025-11-17
Mic crashes when number of channels > 32


b-ma:
test with blackhole 64ch

orottier:
It's probably due to the hard limit on the Rust side

```
/// Maximum number of channels for audio processing
pub const MAX_CHANNELS: usize = 32;
```

This limit has been put in place so we can keep audio buffers (specifically, the `AudioRenderQuantum`) fully on the stack and not add some indirection to the heap.
Increasing the limit will hit performance both in CPU and RAM usage. Also what should the new limit be, 64, 256? There is no clean solution there.

Perhaps changing to a `SmallVec` that keeps buffers up to 32 channels (or less) on the stack and spills to the heap when there is more would be fine. It would have a small CPU usage impact for all users (also the ones with small number of channels). We could benchmark this.

Probably we should move this issue to https://github.com/orottier/web-audio-api-rs/

b-ma:
Hey,

Just to explain a bit more (I just threw this to not forget, but it was a bit too laconic :). I didn't try to grab more than 32 channels with the mic, it was just the fact that more input channels were available from the virtual soundcard that made the script crash. I will try to recreate the setup, will be more clear.

For now, I think this would be ok to just clamp to max number of channels.

PR #194, opened 2026-08-06
Fix - more robust param coercion for MediaDeviceConstraints

### LINKED ISSUE #193: getUserMedia panics ('failed to initiate panic, error 5, aborting') when deviceId constraint is used (macOS)
### Environment
- `node-web-audio-api`: 2.1.0
- OS: macOS (arm64 / Apple Silicon)
- Node: v26.5.0

### Bug
`mediaDevices.getUserMedia()` crashes the whole Node process with a Rust double-panic whenever the `audio` constraints include a `deviceId` (either `exact` or `ideal`), regardless of which input device is targeted:

```
fatal runtime error: failed to initiate panic, error 5, aborting
```

This is not a catchable JS error/rejection — the process aborts immediately.

### Repro

```js
import { mediaDevices } from 'node-web-audio-api';

const devices = await mediaDevices.enumerateDevices();
const target = devices.find(d => d.kind === 'audioinput'); // any input device

// Crashes:
await mediaDevices.getUserMedia({
  audio: { deviceId: { exact: target.deviceId } }
});
```

Also crashes with `{ deviceId: { ideal: target.deviceId } }`, and with any of the 11 input devices I tested (built-in mic, USB webcam mic, Pro Tools Audio Bridge channels, an aggregate device, a Metric Halo MHLink device).

### What does *not* crash

```js
// OK — returns a MediaStream on the default input device
await mediaDevices.getUserMedia({ audio: true });

// OK — channelCount alone works too (though per #74 it may be ignored)
await mediaDevices.getUserMedia({ audio: { channelCount: { ideal: 2 } } });
```

So the crash is specifically tied to the `deviceId` constraint code path, not to `getUserMedia` in general.

### Expected behavior
Either the requested device is opened, or (if the id is invalid/unsupported) the promise rejects with a normal error — not a process-aborting panic.

### Possibly related
- #178 — `getUserMedia`/`enumerateDevices` are not actually async (no napi `Task`), which may explain why a panic during device selection can't be converted into a clean JS rejection.
- #74 — `channelCount` constraint reportedly ignored, suggesting the constraint-handling path for `getUserMedia` is generally unfinished.

Happy to provide more repro details/logs if useful.

b-ma:
Fix #193

### DIFF
--- a/examples/microphone.js
+++ b/examples/microphone.js
@@ -22,10 +22,12 @@ const devices = await mediaDevices.enumerateDevices();
 console.log(devices.filter(d => d.kind === 'audioinput'));
 const inputId = await rl.question('> Input deviceId (empty for default): ');
 
-const mediaStream = await mediaDevices.getUserMedia({ audio: {
-  deviceId: inputId.trim(),
-  channelCount: 33, // more than one channel leads to cranky sound for now...
-}});
+const mediaStream = await mediaDevices.getUserMedia({
+  audio: {
+    deviceId: inputId.trim(),
+    channelCount: 32, // more than one channel leads to cranky sound for now...
+  },
+});
 
 const audioContext = new AudioContext();
 await audioContext.resume();
@@ -33,4 +35,6 @@ await audioContext.resume();
 // const source = new MediaStreamAudioSourceNode(audioContext, { mediaStream });
 const source = audioContext.createMediaStreamSource(mediaStream); // factory API
 source.connect(audioContext.destination);
+
+console.log('> MediaStreamSourceNode number of channels', source.channelCount);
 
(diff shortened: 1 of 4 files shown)


orottier:
Before you release. I am planning a release for the base lib this weekend. Might want to include it, mostly performance and bug fixes, no API changes

b-ma:
ok thanks!

b-ma:
For information I plan to have a look at these points while I'm on this

- [ ] clamp input to 32 channels (cf. #159)
- [ ] report the MediaStreamTrack number of channels
"The number of channels of the output corresponds to the number of channels of the MediaStreamTrack."
https://webaudio.github.io/web-audio-api/#MediaStreamAudioSourceNode
https://webaudio.github.io/web-audio-api/#MediaStreamTrackAudioSourceNode

Hopefully I'll be able to propose something tomorrow that we can include in the release too (otherwise it will probably have to wait for September :)

PR #200, opened 2026-09-14
Chore - drop node-fetch in favor of global fetch

wine-fall:
### What

`node-fetch` is imported in exactly one place — `js/AudioWorklet.js` — to load worklet modules from `http(s)` URLs:

```js
import fetch from 'node-fetch';   // line 17, the only reference in the package
...
const res = await fetch(moduleUrl);   // line 55, addModule() from an http URL
const res = await fetch(url);         // line 79, the wpt caller-site branch
```

Node has shipped a global `fetch` since v18, and `package.json` already declares `"engines": { "node": ">= 22" }` — so the import is dead weight. This PR deletes it and removes the dependency. **The two call sites are unchanged**; nothing else in the diff.

### Why

It removes 5 packages from every install of `node-web-audio-api`:

```
node-fetch
├── data-uri-to-buffer
├── fetch-blob
│   └── node-domexception   ← deprecated
└── formdata-polyfill
```

`node-domexception@1.0.0` is deprecated upstream ("Use your platform's native DOMException instead"), so every `npm i` of a package that depends on this one prints a deprecation warning that the user can do nothing about. That is how I found it — it shows up when installing my own project.

### Behaviour

Both call sites already wrap the fetch in `try/catch` and rethrow a `DOMException('...', 'AbortError')`, so the difference in error shape between `node-fetch` and undici's `fetch` is not observable to callers. Rejection on network failure and `res.text()` behave the same.

### Testing

`node --check js/AudioWorklet.js` passes, and I read every reference to confirm line 17 was the only one (`grep -rn node-fetch` over the tree is now empty). I did not run the WPT suite locally — that needs a full Rust build of the native module on my machine — so I'm relying on CI here. Happy to adjust if something in the worklet tests disagrees.


b-ma:
Hey,

Thanks for the fix! (I checked wpt, no problem)

